PR #4, opened 2018-01-24
Add "go get" install to README

jarongao:
Hello @cosmtrek,

I ran `go get github.com/cosmtrek/air`, and golang automatically loads the `air` binary for use.

I think this is a worthy addition to the Install section of the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,12 @@ chmod +x ~/.air
 ### on Linux
 
 ```bash
+go get github.com/cosmtrek/air # Then run "air" in project directory
+```
+
+or
+
+```bash
 curl -fLo ~/.air \
     https://raw.githubusercontent.com/cosmtrek/air/master/bin/linux/air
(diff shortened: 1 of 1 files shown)


cosmtrek:
Thanks your advice, but I don't think it's a better way to install `air`.

frederikhors:
> Thanks your advice, but I don't think it's a better way to install `air`.

Can I ask you why, @cosmtrek?

Anyway thanks a ton for air.

cosmtrek:
@frederikhors It's better to use the released binary not the code on master branch :)

mingrammer:
@cosmtrek But, you can fix the (stable) version with git tags.

cosmtrek:
@mingrammer Seems good :)

cosmtrek:
@mingrammer @frederikhors Done, you can `go get` to download air now:)

mingrammer:
Great!! Thank you :)

Issue #6, opened 2018-02-27
Address already in use with docker-compose


lnlwd:
Hey, I'm trying to use air within a docker-compose project. The project uses gin as the web server.

Here's my terminal output:
```bash
app_1      | [18:52:21.637] [GIN] 2018/02/27 - 18:52:21 | 200 |      40.743µs |      172.18.0.1 |  GET     /status
app_1      | 
app_1      | [18:52:30.303] main.go has changed
app_1      | [18:52:30.303] building...
app_1      | [18:52:30.875] running...
app_1      | [18:52:30.877] [GIN-debug] [WARNING] Running in "debug" mode. Switch to "release" mode in production.
app_1      |  - using env:	export GIN_MODE=release
app_1      |  - using code:	gin.SetMode(gin.ReleaseMode)
app_1      | 
app_1      | 
app_1      | [18:52:30.877] [GIN-debug] GET    /status                   --> main.main.func1 (3 handlers)
app_1      | [GIN-debug] Listening and serving HTTP on :5000
app_1      | 
app_1      | [18:52:30.877] [GIN-debug] [ERROR] listen tcp :5000: bind: address already in use
```

lnlwd:
Looks like fresh kills the process before starting a new one. With fresh it works.

```bash
app_1      | 19:16:27 runner      | Killing PID 372
```

cosmtrek:
Could you give me more info to reproduce it? Thanks!

lnlwd:
This is my `Dockerfile`:
```dockerfile
FROM golang:1.10-alpine3.7

RUN apk update \
    && apk add git \
    && go get github.com/cosmtrek/air

ENTRYPOINT ["/go/bin/air"]
```

And that's my `docker-compose.yml`:
```yaml
version: '3'

services:
  app:
    build: .
    volumes:
      - .:/go/src/github.com/1STi/core.startupin.me
    working_dir: /go/src/github.com/1STi/core.startupin.me
    ports:
      - "5000:5000"
```

With fresh i got it working with almost the same `Dockerfile`:
```dockerfile
FROM golang:1.10-alpine3.7

RUN apk update \
    && apk add git \
    && go get github.com/pilu/fresh

EXPOSE 5000

ENTRYPOINT ["/go/bin/fresh"]
```

cosmtrek:
@lnlwd Thanks, I'll take time to look at it :)

cosmtrek:
[...]

PR #9, opened 2018-05-11
fix: kill process on darwin

goenning:
Same problem described on #6 happens when running on macOS. 

The code is the same as `utils_linux.go` and it has been tested locally.


cosmtrek:
Wow, thanks!

Issue #8, opened 2018-04-07
Overrides the text color that server might want to render


owais:
A basic hello world example of Gin with and without air

With air:

![image](https://user-images.githubusercontent.com/46186/38453880-9e3be72c-3a7a-11e8-9077-bcd9636da736.png)

Without air

![image](https://user-images.githubusercontent.com/46186/38453891-beb04f98-3a7a-11e8-8fe8-5f1bda1daeda.png)

frederikhors:
Same story here, any news?

cosmtrek:
I'm working on it. But it'll only work on Linux or macOS, Windows is not supported.

cosmtrek:
@owais @frederikhors Hi, I've finished it and released a new version 1.10. Could you try this?

frederikhors:
On Windows same problem also with v1.10.

cosmtrek:
@frederikhors Yes, Windows is not supported now. The point is [pty](https://github.com/kr/pty) does not support Windows.

cosmtrek:
The new version has fixed it.

PR #10, opened 2018-05-11
feature: log configuration to customize log output

goenning:
This PR introduces two minors changes to the logger:

**Change 1:** Skip the line feed if the message already ends with one. The problem is that when the running program outputs a log that ends with a line feed, Air is doubling it. This can also be seen here #8 and my attached screenshot. 

**Change 2:** Introduces a new configuration section to customize log messages. Most applications already include time on its log, so it'd be cleaner if we could just ignore this on Air to prevent double time being logged. This can also be seen here #8 and my attached screenshot. 

Bofore PR:

screen shot 2018-05-11 at 20 00 29

After PR:

screen shot 2018-05-11 at 19 59 56

### DIFF
--- a/air.conf.example
+++ b/air.conf.example
@@ -13,6 +13,9 @@ include_ext = ["go", "tpl", "tmpl", "html"]
 exclude_dir = ["assets", "tmp", "vendor", "frontend/node_modules"]
 delay = 800 # ms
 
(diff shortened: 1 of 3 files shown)


cosmtrek:
You have my gratitude.

PR #22, opened 2019-06-25
Add support exclude file

moonug:


cosmtrek:
Thanks for your work, I've implemented it.

PR #24, opened 2019-08-02
Fixed: old binary will be executed when the building is failure.

caeret:


cosmtrek:
oops... it's a feature not bug 😄

caeret:
well, the new binary will not be generated as expected when there is an error, and the old binary will be executed without be noticed sometimes.

cosmtrek:
2019-11-10 at 21 19 44

The error of build should be displayed as expected. Could you try the new version? Thanks!

PR #30, opened 2019-11-12
Add version flag

rabingaire:
I was on a weird situation, I use make target to install air, after installation I run this command 
`/path/to/air -v > /dev/null` to check if the installation has been successful, but since air did not had -v command I added it in this pull request.


cosmtrek:
Thanks your great work!

PR #43, opened 2020-01-17
Update e.config.Root to get the Absolute path first

spiffyjr:
Resolves the absolute directory for `root` prior to running.


spiffyjr:
This isn't quite working right either. Closing.

PR #47, opened 2020-02-05
Allow specifying list of environment variables.

swhite24:
Thanks for the great project.  I've recently come across a case where I have many environment variables I'd like to provide to the binary and listing them all inline with the `full_bin` key was becoming a bit unwieldy.

This PR adds a new key, `env`, which will accept a table of inputs to prepend to the bin command as environment variables.

For example, this config (truncated):

```toml
# air.conf
[build]
bin = "tmp/main"
env.PORT = "3000"
env.ENV = "production"
```

Will run the following:
```
PORT=3000 ENV=production tmp/main
```

### DIFF
--- a/air_example.conf
+++ b/air_example.conf
@@ -10,8 +10,11 @@ tmp_dir = "tmp"
 cmd = "go build -o ./tmp/main ."
 # Binary file yields from `cmd`.
 bin = "tmp/main"
-# Customize binary.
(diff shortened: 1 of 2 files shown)


udaya2899:
Wouldn't it be better to ask the user to have the list of env files as export commands in a separate file called `dev.env` and source the file before building, with bash it'd be `source dev.env`

PR #48, opened 2020-02-09
Use mergo for config loading

momotaro98:
## What does this do / why do we need it?

This PR resolves [one of the TODOs](https://github.com/cosmtrek/air/blob/ef30fc85f914b4bb8615f5ea875dba02c3f36cb7/runner/config.go#L123), `// TODO: maybe better way to assign`.

## How this PR fixes the problem?

I adopt [mergo](https://github.com/imdario/mergo), which has a perfect function for the TODO.

Also, I did small refactoring in `initConfig` func.

## Test case

[x] `make check`
[x] `make build`

As far as I can, I confirmed this RR doesn't break existing functions.

I checked 
* if the config in toml file is applied
* if default is applied when toml file has empty field
* if handle `-c` option is set or not


cosmtrek:
mergo is really great and thanks for your contribution!

PR #51, opened 2020-02-12
KillWait

### LINKED ISSUE #29: feature-request：About the signal when the killCmd method is executed
🎉Thank you for your air, I like it very much.

I use Gin to write code for the server.
When the server will be shut down，I need to do some processing like closing the database connection.
But because the killCmd method sent the syscall.SIGKILL signal, there is no time to close the database connection.
When my Gin server is started again by air, an error can occur because the database connection is not closed properly and the server cannot be started.
Can I configure the semaphore in config so that I can pass syscall.SIGTERM.
It is better if air can set a timeout for waiting for the kill command to end.

This is just a suggestion, I hope to improve the air this useful tool.

ghost:
Fixes #29 
Added option to wait for kill of previous version


runner/engine.go
```diff
@@ -324,12 +339,16 @@ func (e *Engine) runBin() error {
...
-		if err = os.Remove(cmdBinPath); err != nil {
-			e.mainLog("failed to remove %s, error: %s", e.config.rel(e.config.binPath()), err)
+
+		if e.config.Build.KillWait == false {
+			// This is already a new bin, do not need to delete it
+			cmdBinPath := cmdPath(e.config.binPath())
+			if _, err = os.Stat(cmdBinPath); os.IsNotExist(err) {
+				return
+			}
+			if err = os.Remove(cmdBinPath); err != nil {
```
wsnotify: Do we even need to delete bin file here? We already have a `cleanup()`

wsnotify: @cosmtrek what do you think?

PR #64, opened 2020-04-16
Worn if tmp is dot

musinit:
Also added warning log with red default color

### DIFF
--- a/air_example.conf
+++ b/air_example.conf
@@ -41,6 +41,7 @@ main = "magenta"
 watcher = "cyan"
 build = "yellow"
 runner = "green"
+warning = "red"
 
 [misc]
 # Delete tmp directory on exit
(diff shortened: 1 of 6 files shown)


udaya2899:
Are PR's actually accepted or responded with, here?

musinit:
Looks like they don't accepted for a while

xiantang:
i think use warning is a not better way. because you also warning this message and use know this problem they use ctrl + c to stop the process and his project Always will be clean. as I think you should raise a err and stop the project!

xiantang:
@musinit

xiantang:
pr is welcome!

HeCorr:
I agree with @xiantang. Data loss should not be acceptable at all. Instead of a warning, the best option is to print an error message and exit the application immediately, without triggering the cleanup.

PR #79, opened 2020-07-02
:sparkles: Added support for custom shell

### LINKED ISSUE #77: Option to use bash instead of sh
I have my environment variables in a file called `dev.env` which I'd want to `source` by executing `source dev.env` before building, since the cmd command is executed inside a `sh` and not `bash`, I cannot source the file.

I tried doing the following

`cmd = ". ./dev.env && go build -o ./tmp/main ."`

But it doesn't work. The variables exported in dev.env doesn't impact the project.
It'd be great if cmd command was executed using `bash` inside air and all this would work like usual I believe

INFO: I am running air in bash shell, dev.env is in root of project directory

TL;DR Need bash support

udaya2899:
## Gives the user, option to use a custom bash shell instead of executing with default "sh" hardcoded. 
Fixes #77 

Now, add
`shell = <shell_name>` in .air.conf file and air builds using the specific shell.
Example,
`shell = bash`
`shell = dash`
`shell = zsh`

`cmd` runs in the respective shell name specified.

Use case:
1. If user has a custom environment variables as a .env file and wants to source it every time before building. If user uses `bash` for development locally, can add `shell = bash` in .air.conf and the command specified as `cmd = ...` will execute as `/bin/bash` and not as `/bin/sh`

My use case: I used `bash` to source a file with command `source dev.env` and `source` doesn't work in `sh`, the alternative didn't work too as explained in #77

### DIFF
--- a/runner/config.go
+++ b/runner/config.go
@@ -28,6 +28,7 @@ type config struct {
 
 type cfgBuild struct {
 	Cmd           string        `toml:"cmd"`
+	Shell         string        `toml:"shell"`
 	Bin           string        `toml:"bin"`
 	FullBin       string        `toml:"full_bin"`
 	Log           string        `toml:"log"`
@@ -86,6 +87,7 @@ func initConfig(path string) (cfg *config, err error) {
 func defaultConfig() config {
 	build := cfgBuild{
 		Cmd:         "go build -o ./tmp/main .",
+		Shell:       "sh",
 		Bin:         "./tmp/main",
 		Log:         "build-errors.log",
 		IncludeExt:  []string{"go", "tpl", "tmpl", "html"},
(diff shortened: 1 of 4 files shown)


udaya2899:
Hello @cosmtrek 
Can you check this out

udaya2899:
@xiantang  Can you please verify this, I'd like to share the project with colleagues, please refer to the issue and my PR. It has been stale for long. Thanks in advance

xiantang:
@udaya2899 i will verify this soon. and I think we maybe not need to add a new Field `shell`. it maybe cause use to config more. we can talk about this.

udaya2899:
I can understand the overhead it creates on each additional config variable. Please suggest alternatives if available. I'll want to use `bash` instead.

PR #82, opened 2020-07-07
use filepath.Match to support shell file name pattern

sprhawk:
I am using emacs, which will create temporary file as .#xxxx.go, so I need air to support shell pattern, like ".#*", "#*", etc in exclude file. I use filepath.Match to support it.

### DIFF
--- a/runner/util.go
+++ b/runner/util.go
@@ -116,8 +116,13 @@ func (e *Engine) isIncludeExt(path string) bool {
 func (e *Engine) isExcludeFile(path string) bool {
 	cleanName := cleanPath(e.config.rel(path))
 	for _, d := range e.config.Build.ExcludeFile {
-		if d == cleanName {
-			return true
+		dir := filepath.Dir(cleanName)
+		pattern := filepath.Join(dir, d)
+		matched, err := filepath.Match(pattern, cleanName)
+		if err == nil {
+			if matched == true {
+				return true
+			}
 		}
 	}
 	return false


xiantang:
@sprhawk  please provide unit tests

silverwind:
I think both `include_dir` and `exclude_dir` should be able to take glob patterns. Also would add some examples to `air_example.conf` so that feature is easy to discover.

xiantang:
we can support Regular Expression.and we can reduce `include_dir` and `include_file` in one Field

sprhawk:
> I think both `include_dir` and `exclude_dir` should be able to take glob patterns. Also would add some examples to `air_example.conf` so that feature is easy to discover.

I just put my PR as a proposal, I'm not sure whether you like to add patterns to include_* or exclude_* ( may be you have some consideration )

sprhawk:
> > I think both `include_dir` and `exclude_dir` should be able to take glob patterns. Also would add some examples to `air_example.conf` so that feature is easy to discover.
> 
> We can support Regular Expression.and we can reduce `include_dir` and `include_file` in one field.

Sure we can use regexp, but what I considered is glob pattern is simpler to use. for most of cases, simple glob pattern is good enough.

I think if being willing to add regexp support, maybe it should have separated configurations, eg `exclude_regex`, `include_regex`,  is more appropriate.

sprhawk:
> @sprhawk please provide unit tests

Hi, I was busy doing project. Do you still need a unittest to merge PR or will you make your own changes later?

PR #88, opened 2020-07-23
create github action workflow and add IsDir() test case

xiantang:
try to fix https://github.com/cosmtrek/air/issues/87

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -34,4 +34,4 @@ jobs:
       run: go build -v .
 
     - name: Test
-      run: go test -v .
+      run: go test -v ./...
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -13,22 +13,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
-    - name: Set up Go 1.x
+    - name: Set up Go 1.4
       uses: actions/setup-go@v2
       with:
-        go-version: ^1.13
+        go-version: ^1.14
       id: go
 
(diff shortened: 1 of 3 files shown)


xiantang:
try to fix https://github.com/cosmtrek/air/issues/87

.github/workflows/go.yml
```diff
@@ -0,0 +1,37 @@
...
+        go-version: ^1.13
+      id: go
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Get dependencies
+      run: |
+        go get -v -t -d ./...
+        if [ -f Gopkg.toml ]; then
```
cosmtrek, line 28: Air is using `go mod` to manage dependencies, so it's unnecessary to check if using dep.

.github/workflows/go.yml
```diff
@@ -0,0 +1,37 @@
...
+
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Set up Go 1.x
+      uses: actions/setup-go@v2
+      with:
+        go-version: ^1.13
```
cosmtrek, line 19: Can we use go 1.14?

xiantang:
@cosmtrek  we need add required must 1 reviewer rule in some create a pull request.

.github/workflows/go.yml
xiantang: we can!

.github/workflows/go.yml
xiantang: done

PR #90, opened 2020-07-25
Fix configuration file extension from `conf` to `toml`

178inaba:
The configuration file is written in TOML format.
https://github.com/cosmtrek/air/blob/66e5eae8b270ea98ac50ac85f355d30c8607e1ca/air_example.conf#L1

So, by making the extension `toml`, you can apply syntax highlighting in TOML format and it will be easier to see.
https://github.com/cosmtrek/air/blob/3c279bf06a7243437e04b7f7c1ac679c63b123c7/air_example.toml


runner/config.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	dftConf = ".air.conf"
+	dftConf = ".air.toml"
```
xiantang, line 16: We need to consider compatibility problem.if we accept this alter. and user update the air,will cause the air not work.

xiantang:
we can both support old style use conf and support new style use toml.

xiantang:
@178inaba

runner/config.go
178inaba: @xiantang 
Fix to support both `.air.toml` and `.air.conf`!

runner/config.go
```diff
@@ -83,6 +78,30 @@ func initConfig(path string) (cfg *config, err error) {
 	return cfg, err
 }
 
+func defaultPathConfig() (*config, error) {
+	// when path is blank, first find `.air.toml`, `.air.conf` in `air_wd` and current working directory, if not found, use defaults
+	for _, name := range []string{dftTOML, dftConf} {
+		var path string
+		if wd := os.Getenv(airWd); wd != "" {
```
xiantang, line 85: if user use dftConf. can we print a warning log to tell user, .conf will be Deprecated soon, recommend to use .toml .

xiantang:
i will verify this pr soon.

runner/config.go
178inaba: @xiantang Added!
https://github.com/cosmtrek/air/pull/90/commits/9216bb2d99ac3603ec93be9d6944ab6e7feb0055

xiantang:
LGTM

xiantang:
thank you!

PR #92, opened 2020-07-25
Place the built binary on the release page

### LINKED ISSUE #75: Binaries shouldn't be committed to VCS
Binaries generally shouldn't be committed to VCS. I recommend creating a GitHub action that automatically generates the binaries and attaches them to a release.

178inaba:
Using [GoReleaser Action](https://github.com/goreleaser/goreleaser-action).

Fixes #75.


.github/workflows/go.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
-    - name: Set up Go 1.4
+    - name: Set up Go 1.14
```
xiantang, line 16: ops! my fault

.github/workflows/go.yml
```diff
@@ -27,3 +27,12 @@ jobs:
 
     - name: Test
       run: go test -v ./...
+
+    - name: Run GoReleaser
```
xiantang, line 31: let me test this. I will rely soon

README.md
```diff
@@ -36,23 +36,9 @@ The classic way to install
 go get -u github.com/cosmtrek/air
 ```
 
-### macOS
+### macOS, Linux, Windows
 
-```bash
```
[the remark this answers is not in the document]
178inaba: @xiantang 
The URL is redirected to the master's bin directory.

```
$ curl -i https://git.io/darwin_air
HTTP/1.1 302 Found
...
Location: https://raw.githubusercontent.com/cosmtrek/air/master/bin/darwin/air
...
```

The issue of #75 had a problem with committing binaries, so I deleted the bin directory.
So if execute `curl -fLo air https://git.io/darwin_air`, will be redirected to a non-existing binary and will get an error.

.github/workflows/go.yml
178inaba: @xiantang OK!

xiantang: it's not work.  because this job on other pull request or push. we can create a new workflow to do this.
```
ame: goreleaser

on:
  pull_request:
  push:

jobs:
  goreleaser:
    runs-on: ubuntu-latest
    steps:
      -
        name: Checkout
        uses: actions/checkout@v2
        with:
          fetch-depth: 0
      -
        name: Set up Go
        uses: actions/setup-go@v2
        with:
          go-version: 1.14
      -
        name: Run GoReleaser
        uses: goreleaser/goreleaser-action@v2
        with:
          version: latest
          args: release --rm-dist
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

xiantang: you can separate this part , and i will receive other change in this pull request.

178inaba: @xiantang 
I added release workflow to GitHub Actions!
https://github.com/cosmtrek/air/pull/92/commits/3463fccbb0842613e9c867724d45aa0ce2937554

178inaba: Another way.
There is a way to add it as a separate job to an existing workflow.
This allows the job to run only when the tag is pushed by using the conditional statement (`if: startsWith(github.ref, 'refs/tags/v')`).
https://github.com/cosmtrek/air/pull/92/commits/82d58e3e89e7ed6f148ba78830dc15b8259d5b93

Which would you like to choose?
If you decide which method to use, I will delete the one that you do not use.

xiantang: @178inaba I test this. it's work! After thinking,I think I want decide choose your recommend way just merge this release as a job. because build job pass is necessary before release.so separate a new file to release is not good. you delete the one I not use. and I will merge this feature to master.

178inaba: @xiantang 
OK!
I have deleted the release workflow files!
https://github.com/cosmtrek/air/pull/92/commits/32e83bbf0e2acf0a869574ed28fcdb4eed72bfcf

PR #95, opened 2020-08-02
feat(*): add golangci-lint checker

cosmtrek:


.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,30 @@
+name: golangci-lint
+
+on:
+  push:
+    tags:
+      - v*
+    branches:
+      - master
```
xiantang, line 8: can we set this on all branches?

cosmtrek: OK

cosmtrek:
@xiantang PTAL

PR #98, opened 2020-08-07
ci: Add archives.format setting to .goreleaser.yml

tkms0106:
For the purpose of allowing users to download binary directly in release page

cf. https://goreleaser.com/customization/archive/


xiantang:
i will check it soon

xiantang:
LGTM

PR #118, opened 2020-10-20
improve on windows

tyahha:
This PR improve two things on Windows.

The first thing is fail to kill process(configured by "full_bin").
Before this PR, using "start" command (windows native command) for start process.
The "start" command lunch another child process.
air recognize "start" command's PID but not "start command"'s child process PID.
Therefore fail to kill process.

The second thing is improvement "full_bin" configuration on Windows.
Before this PR. using "cmd" command (windows native command) for start process.
As the "cmd" command is native windows command, this command require '\' for path separator.
So We can't use '/' for path separator.
this was improved by not using "cmd", but calling directly "full_bin" configuration's command.

### DIFF
--- a/runner/util.go
+++ b/runner/util.go
@@ -182,7 +182,6 @@ func adaptToVariousPlatforms(c *config) {
 	// Use the unix configuration on Windows
 	if runtime.GOOS == PlatformWindows {
 
-		runName := "start"
 		extName := ".exe"
 		originBin := c.Build.Bin
 		if !strings.HasSuffix(c.Build.Bin, extName) {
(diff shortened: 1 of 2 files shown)


tyahha:
following PR improve this PR's problem.
And following PR is simple and hasn't this PR's "full_path" configuration's problem.
This PR parse "full_path" with strings.Split(separator is space).
So when execution, "full_path" have no space.
But, it also removes the necessary space.
For example, spaces in a string.
Therefore, close this PR.

https://github.com/cosmtrek/air/pull/119

PR #121, opened 2020-11-12
feat: Skip write events for unchanged files

bbrks:
Can be enabled with the `[build].exclude_unchanged = true` config option. This is disabled by default.

Editors often save a file even if the contents of the file hasn't changed.
We get the fsnotify event even though there is nothing to be rebuilt in this case.

This PR allows `air` to only rebuild projects if the contents of the files have actually changed.

On startup, we calculate (sha256) checksums of all the watched files, and store them in a cache.
After a file write event, the file contents is hashed to compare it to the previously cached checksum. If the checksums match, the write event is skipped, otherwise the cache is updated and the write event continues.

### Running with unchanged file save, then modification:
Screenshot_20201112_102918

### Startup and file change with debug:
Screenshot_20201112_102730
Screenshot_20201112_102833

### DIFF
--- a/air_example.toml
+++ b/air_example.toml
@@ -20,6 +20,8 @@ exclude_dir = ["assets", "tmp", "vendor", "frontend/node_modules"]
 include_dir = []
 # Exclude files.
 exclude_file = []
+# Exclude unchanged files.
+exclude_unchanged = true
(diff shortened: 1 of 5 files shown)


xiantang:
LGTM

PR #137, opened 2021-02-02
Follow symlink if required

zhengcan:
Changes:
- add a new option: `follow_symlink`
- able to follow symlink to scan and watch dirs


runner/config.go
```diff
@@ -37,6 +37,7 @@ type cfgBuild struct {
 	IncludeDir       []string      `toml:"include_dir"`
 	ExcludeFile      []string      `toml:"exclude_file"`
 	ExcludeUnchanged bool          `toml:"exclude_unchanged"`
+	FollowSymlink    bool          `toml:"follow_symlink"`
```
cosmtrek, line 40: It's better to add usage in `air_example.toml`.

zhengcan: It has been updated.

cosmtrek (approved):

PR #148, opened 2021-03-06
Fix process not being killed on Windows

### LINKED ISSUE #147: Process not being killed on Windows
Hello! First of all, thank you for this awesome package  :)

Due to Air's current implementation of starting processes on Windows (with `start /b`), the PID returned by `cmd.Process.Pid` is of a process that already exited (`start`'s PID, not what it started), hence the debug error `failed to kill PID ###, error: exit status 128` is returned.

When trying to run the `TASKKILL` command manually, I noticed the following error: `ERROR: The process "###" was not found.`, which confirms my suspicion.

I'm thinking of a good way to fix this, since just `TASKKILL`ing the process by name isn't a good idea.

Is `start /b` even necessary? I'll try to modify the code and run some tests without it.

If that doesn't work, maybe try to get the PID by executable path? Is that even possible? If so, just check it against `c.Build.FullBin` and use that on `TASKKILL`.

I don't think it would be a good idea to try to parse `TASKLIST` as the language might change per system, etc. (also it doesn't even seem to return the executable path).. but I don't see other alternatives as listing processes in Go [seems to be a nightmare](https://stackoverflow.com/a/46849516)..

Suggestions?

HeCorr:
Closes #147

### DIFF
--- a/main.go
+++ b/main.go
@@ -61,5 +61,4 @@ func main() {
 	}()
 
 	r.Run()
-	<-r.ExitChan()
 }
--- a/main.go
+++ b/main.go
@@ -61,4 +61,5 @@ func main() {
 	}()
 
 	r.Run()
+	<-r.ExitChan()
 }
--- a/runner/engine.go
+++ b/runner/engine.go
@@ -22,6 +22,7 @@ type Engine struct {
 	watcherStopCh  chan bool
 	buildRunCh     chan bool
 	buildRunStopCh chan bool
+	canExit        chan bool
 	binStopCh      chan bool
 	exitCh         chan bool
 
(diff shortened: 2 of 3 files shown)


HeCorr:
Sorry, I just noticed this causes a panic when reloading for the second time.. Lemme fix it..

HeCorr:
Should be good now! I tested multiple scenarios..

- Reloading once
- Reloading more than once
- Closing after starting
- Closing after reloading once
- Closing after reloading more than once

HeCorr:
aaaaaand I just found another panic... please don't merge it yet..

anton7r (approved):
This fixes the issue

HeCorr:
I fixed the panic that was happening when the build failed twice in a row.

(please `git pull` from my fork if you had cloned it and `go install` again to have yours patched.)

@anton7r / @tusharsadhwani / @EnzoTools

anton7r:
Sorry for being late to check out the changes @HeCorr!

I've tested the new patch and it seems to work really well, good job! 👏

anton7r (approved):

xiantang (approved):

HeCorr:
thank you very much!

PR #158, opened 2021-04-17
ExcludeRegex

KyleU:
Adding `ExcludeRegex` to build config, excluding files based on compiled regular expressions. Test included.

### DIFF
--- a/air_example.toml
+++ b/air_example.toml
@@ -20,6 +20,8 @@ exclude_dir = ["assets", "tmp", "vendor", "frontend/node_modules"]
 include_dir = []
 # Exclude files.
 exclude_file = []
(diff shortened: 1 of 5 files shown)


KyleU:
Thanks for building `air`, I've used it in a dozen projects. This PR covers the case where you want to exclude files based on a regex pattern. In my case, I was using quicktemplate to generate `.html.go` files from `.html`, and needed to exclude only the generated files in that directory.

KyleU:
@xiantang Should I open an issue for this, or is the PR sufficient?

cosmtrek:
@KyleU Thanks for your contributions, I'll review it~

cosmtrek (approved):

Issue #153, opened 2021-03-18
I want to output .air.toml from cli.


dohq:
Hi.
Thank you for the great useful tool!

 it might be useful to be able to output the sample configuration file from the command.
This tool works broadly ideal even without a configuration file.
So, in the rare case that you need a configuration file,
you will need to copy and paste the file from github every time or remember which project the existing configuration file was used in.
Fortunately, they seem to be using go1.16 these days, so I used embed to embed air_example.toml.
https://github.com/dohq/air/commit/a1cffeef80815764598833f4dbbc0aa6aa50ca98

If you are okay with adding this feature, create a pull request.
Cheers!

cosmtrek:
@dohq Seems great, could you create a pr?

PR #159, opened 2021-04-20
Add init command to create a .air.toml file to the current directory.

lime008:
Calling `air init` will create the .air.toml based on the default configuration to the current directory. 

This will make it easier to start using a custom config in a new project.

Resolves: #153

### DIFF
--- a/runner/config.go
+++ b/runner/config.go
@@ -80,7 +80,7 @@ func initConfig(path string) (cfg *config, err error) {
 	return cfg, err
 }
 
-func writeConfig() {
+func writeDefaultConfig() {
 	confFiles := []string{dftTOML, dftConf}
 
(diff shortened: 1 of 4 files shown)


cosmtrek:
@lime008 Looks good, please  update REAME to show how to use it.

runner/config.go
```diff
@@ -79,6 +80,41 @@ func initConfig(path string) (cfg *config, err error) {
 	return cfg, err
 }
 
+func writeConfig() {
```
cosmtrek, line 83: rename to `writeDefaultConfig`

lime008:
Updated the README. Do you think this is sufficient? @cosmtrek

cosmtrek (approved):

PR #202, opened 2021-11-08
Exclude node_modules by default

EwenQuim:
Nobody wants having to deal with `node_modules`, and I think it's a nice default.

They are the only reason I used `air init`.

PS: Air is a really nice CLI, I use it in all my projects ! If you do not agree with this PR, no problem, you can refuse it.

### DIFF
--- a/runner/config.go
+++ b/runner/config.go
@@ -171,7 +171,7 @@ func defaultConfig() config {
 		Bin:          "./tmp/main",
 		Log:          "build-errors.log",
 		IncludeExt:   []string{"go", "tpl", "tmpl", "html"},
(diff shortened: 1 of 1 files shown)


servusdei2018:
I see the reasoning behind this, but node isn't alone, some people may have `__pycache__` instead, etc. While the focus is on go, air can be used in polyglot environments too, although I don't think it's the main focus at the moment. Just tossing this idea out there but perhaps we can have air understand and respect a `.gitignore` (or maybe an `.airignore`) in order to separate this kind of logic from air itself.

xiantang:
I think you can put node_modules in to  air_example.toml

PR #204, opened 2021-11-09
Allow linux/arm64 platforms it work with docker over m1

jonafrank:
When Docker runs over darwin/arm64 usually convert the platform to linux/arm64 just added this platform to allow to run over docker


xiantang (approved):

xiantang:
LGTM

PR #229, opened 2021-12-05
io/ioutil package is deprecated

xiantang:


xiantang:
https://github.com/cosmtrek/air/pull/224 @furusax0621  check is running. if smoke test no issue I will merge it.

nwadams:
@xiantang This change bumps the minimum golang version of this project to 1.16. You might need to update the docs to reflect that since it will break for anyone using an older version of go.

PR #233, opened 2021-12-25
Fix interrupt got "no such process"

rpoetrap:


xiantang:
thanks run pipeline first..

runner/util_darwin.go
```diff
@@ -13,7 +12,7 @@ func (e *Engine) killCmd(cmd *exec.Cmd) (pid int, err error) {
 
 	if e.config.Build.SendInterrupt {
 		// Sending a signal to make it clear to the process that it is time to turn off
-		if err = syscall.Kill(-pid, syscall.SIGINT); err != nil {
```
xiantang, line 16: we use -pid is equal to  send interrupt to his pgid
>  Nega-
       tive PID values may be used to choose whole process groups; see the PGID column in ps command output

xiantang: this is in `man kill`

PR #243, opened 2022-01-23
fix cannot kill all child process issue

xiantang:
https://stackoverflow.com/questions/392022/whats-the-best-way-to-send-a-signal-to-all-members-of-a-process-group


xiantang:
close https://github.com/cosmtrek/air/issues/241

PR #251, opened 2022-02-08
Fix kill when building

xiantang:


xiantang:
close https://github.com/cosmtrek/air/issues/183

PR #254, opened 2022-02-22
ensure multiple processes stop to avoid address already in use errors

nwadams:
When multiple builds are started in parallel we can end up in situations where more than one process is running at once. In a web server this leads to `2022/02/21 21:23:24 listen tcp :1234: bind: address already in use` and the appearance that code is not being updated and requires stopping/restarting air and potentially tracking down and killing orphaned processes in order to recover to a good state. 

This PR keeps track of each process started in a map based on PID and maintains a per process channel in order to be able to stop all previous processes when running a build, as well as when running `runBin`. By doing this we avoid the impacts of concurrent builds and prevent orphaned processes. This PR fixes https://github.com/cosmtrek/air/issues/216

### DIFF
--- a/runner/engine.go
+++ b/runner/engine.go
@@ -24,11 +24,10 @@ type Engine struct {
 	buildRunCh     chan bool
 	buildRunStopCh chan bool
 	canExit        chan bool
-	binStopCh      chan bool
+	binStopChMap   map[int]chan bool
 	exitCh         chan bool
 
 	mu            sync.RWMutex
-	binRunning    bool
 	watchers      uint
 	fileChecksums *checksumMap
 
@@ -58,9 +57,8 @@ func NewEngine(cfgPath string, debugMode bool) (*Engine, error) {
 		buildRunCh:     make(chan bool, 1),
 		buildRunStopCh: make(chan bool, 1),
 		canExit:        make(chan bool, 1),
-		binStopCh:      make(chan bool),
+		binStopChMap:   make(map[int]chan bool),
 		exitCh:         make(chan bool),
-		binRunning:     false,
 		watchers:       0,
 	}
 
@@ -342,9 +340,10 @@ func (e *Engine) start() {
 
 		// if current app is running, stop it
 		e.withLock(func() {
-			if e.binRunning {
-				e.binStopCh <- true
+			for _, v := range e.binStopChMap {
+				v <- true
 			}
+			e.binStopChMap = make(map[int]chan bool)
 		})
 		go e.buildRun()
 	}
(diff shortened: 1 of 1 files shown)


nwadams:
@xiantang if this idea makes sense I'll look at adding some test coverage so the codecov passes if you want.

xiantang:
I think this one is makes sense. plz add more coverage here.

runner/engine.go
```diff
@@ -437,7 +440,7 @@ func (e *Engine) runBin() error {
 			default:
 			}
 		}()
-		<-e.binStopCh
+		<-e.binStopChMap[cmd.Process.Pid]
```
xiantang, line 443: need check nil?
bcs when we read from a nil channel it will block forever.

nwadams: I'll look into this. I think I want to make sure our locking ensures we never attempt to release before we start listening so the channel is never nil. 

I think I'll pass in the channel to the function here instead of looking it up on demand to protect against it already having been messaged.

nwadams: I did a little refactoring that should prevent reading from a nil channel.

xiantang:
i'll take a look soon.

nwadams:
@xiantang I improved the codecov, but I'm not actually sure my functionality (or the original functionality) is tested in a unit test. To properly test this code I think we would need hooks to simulate file changes and process reloads in a unit test. This style of testing may be more appropriate for an integration test similar to your existing smoke test to verify that we don't see deadlocks or other bad behavior. 

Thanks for your time taking a look at this and I will look at adding support for eagerly stopping builds if/when multiple start concurrently in a future PR since I think that can be valuable for a larger app.

runner/engine.go
```diff
@@ -472,9 +473,10 @@ func (e *Engine) cleanup() {
 	defer e.mainLog("see you again~")
 
 	e.withLock(func() {
-		if e.binRunning {
-			e.binStopCh <- true
+		for _, v := range e.binStopChMap {
```
xiantang, line 476: can we use close(e.binRunning) ?

xiantang: then recreate a new channel e.binStopCh

runner/engine.go
xiantang: I dont think we need a map. just use close() to broadcast.

xiantang:
https://github.com/cosmtrek/air/pull/255 plz check this branch.

xiantang:
> #255 plz check this branch.

I've use close to notify go goroutines.

xiantang:
LGTM totally. after merge this PR, I will merge https://github.com/cosmtrek/air/pull/255

quinn:
I still have this issue frequently, multiple times a day. Version:

```
  __    _   ___
 / /\  | | | |_)
/_/--\ |_| |_| \_ v1.52.2, built with Go go1.22.4

```

PR #257, opened 2022-02-25
fix interrupt handling on darwin

roccoblues:
We need a negative pid here to kill the whole process group.

### DIFF
--- a/runner/util_darwin.go
+++ b/runner/util_darwin.go
@@ -12,18 +12,18 @@ import (
 
 func (e *Engine) killCmd(cmd *exec.Cmd) (pid int, err error) {
 	pid = cmd.Process.Pid
+	pgid, err := syscall.Getpgid(cmd.Process.Pid)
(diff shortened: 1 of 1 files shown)


runner/util_darwin.go
```diff
@@ -15,7 +15,7 @@ func (e *Engine) killCmd(cmd *exec.Cmd) (pid int, err error) {
 
 	if e.config.Build.SendInterrupt {
 		// Sending a signal to make it clear to the process that it is time to turn off
```
xiantang, line 17:
```
pgid, err := syscall.Getpgid(cmd.Process.Pid)
	if err != nil {
		return pgid, errors.Wrapf(err, "failed to get pgid, pid %v", pid)
	}
```
should put this code in front syscall.Kill
than use -pgid to kill

xiantang (approved):

xiantang:
Merged. Thanks.

PR #268, opened 2022-04-16
📦 NEW: Updated go to 1.18 in Docker

Anthodev:


xiantang:
Why do we need to upgrade to v1.18? if `air` is upgraded to v1.18, it will cause a lot of users unable to use when they use go lower than v1.18

Anthodev:
> Why do we need to upgrade to v1.18? if `air` is upgraded to v1.18, it will cause a lot of users unable to use it when they use go lower than v1.18

It's a proposition to give the possibility for the users to have a 1.18 docker image ready to use (instead of mounting an image from golang 1.18 and install `air` on it) and so to be up to date with go.

Merge it when you want.

xiantang:
> > Why do we need to upgrade to v1.18? if `air` is upgraded to v1.18, it will cause a lot of users unable to use it when they use go lower than v1.18
> 
> It's a proposition to give the possibility for the users to have a 1.18 docker image ready to use (instead of mounting an image from golang 1.18 and install `air` on it) and so to be up to date with go.
> 
> Merge it when you want.

so in this case, just need to change the GO version to v1.18 in the docker, right? no need to change GO version in `go.mod` etc.

DriLLFreAK100:
Good day, @xiantang @Anthodev . Any updates on this? Would be a great help to get the supporting image for 1.18 in. Thanks in advance!

Anthodev:
> Good day, @xiantang @Anthodev . Any updates on this? Would be a great help to get the supporting image for 1.18 in. Thanks in advance!

I made the change two weeks ago, i'm just waiting to be merged :)

desol:
This image would be very helpful for me and my team.

JeremyCraven:
This would be helpful.

PR #282, opened 2022-05-30
Allow to write application logs to file

filipenos:
Hello great project, congratulations.

I thought I'd share this idea, allow writing the application logs in a separate file (stderr.log and stdout.log)

Basically using a io.MultiWriter


xiantang:
I think can use `air | tee output.log`, BTW seems cannot build this code

PR #303, opened 2022-07-02
fixes air_example.toml regex expression

jshnaidman:
I think this regex expression should be "_test\\.go". The '.' character matches any character in regex.

### DIFF
--- a/air_example.toml
+++ b/air_example.toml
@@ -21,7 +21,7 @@ include_dir = []
 # Exclude files.
 exclude_file = []
 # Exclude specific regular expressions.
-exclude_regex = ["_test.go"]
+exclude_regex = ["_test\\.go"]
 # Exclude unchanged files.
 exclude_unchanged = true
 # Follow symlink for directories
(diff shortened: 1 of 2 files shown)


xiantang:
could u add a test for this ? @jshnaidman

jshnaidman:
I'm a little confused that this affects code coverage since I only updated the documentation. I added some test cases which cover the changes I made to the documentation.

xiantang:
> I'm a little confused that this affects code coverage since I only updated the documentation. I added some test cases which cover the changes I made to the documentation.

thanks

air_example.toml
```diff
@@ -21,7 +21,7 @@ include_dir = []
 # Exclude files.
 exclude_file = []
 # Exclude specific regular expressions.
-exclude_regex = ["_test.go"]
+exclude_regex = ["_test\\.go"]
```
xiantang, line 24: `_test\\.go`  should be `_test\.go`

jshnaidman: Without `` the string won't escape the \ so you need to escape it once for the "" string, and then again to escape the . So \\ I think.

jshnaidman: Sorry I didn't realize github didn't escape, I mean't I think it should be \\\\

PR #315, opened 2022-08-01
Make root config support path like "..", "../..", etc

fldm713:
For example, my project tree look like this:
```
Myproject/
┣ example/
┃ ┣ .air.toml
┃ ┣ go.mod
┃ ┣ go.sum
┃ ┗ main.go
┣ module/
┃ ┣ go.mod
┃ ┗ module.go
┣ .gitignore
┗ go.work
```
I want to run `air` in `example` folder but watch `Myproject` folder. In `.air.toml`, if I set root to `..`, it does not work, I have to set it to `../../Myproject`.

This PR makes support for `..` for root setting.

### DIFF
--- a/runner/util.go
+++ b/runner/util.go
@@ -65,7 +65,7 @@ func (e *Engine) isTestDataDir(path string) bool {
 }
 
 func isHiddenDirectory(path string) bool {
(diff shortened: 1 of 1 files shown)


xiantang:
LGTM

PR #318, opened 2022-08-02
Handle absolute paths in config

ZipFile:
Was trying to use `/tmp/air` as `tmp_dir` but it did not work well. `filepath.Join` simply concatenates both components, resulting in something like `/home/user/myproj/tmp/air`, which is counter-intuitive.


ZipFile:
Yep, docker with restricted write permissions was my exact use case. Anyway, rebased with the latest `master` since there is an interest in this PR.

xiantang (commented):
LGTM good test case

PR #322, opened 2022-08-12
add support for pre build commands with extension restriction (pre bu…

corneldamian:
add support for pre build commands with extension restriction

[[build.pre_builder]]
  cmd = "gqlgen generate"
  only_ext = "graphql"
[[build.pre_builder]]
  cmd = "go generate"
  only_ext = "graphql"


xiantang:
https://github.com/cosmtrek/air/issues/453 thanks you so much, but we already support it now

PR #323, opened 2022-08-17
feat(config): only convert binary path to abs

dragonly:
The current implementation will convert the whole config.bin to the abs path. This introduces errors when the parameter contains "//", like "--url http://example.com".

This PR changes to only convert the binary path part (excluding the parameters) into the abs path.

### DIFF
--- a/runner/config.go
+++ b/runner/config.go
@@ -9,6 +9,7 @@ import (
 	"reflect"
 	"regexp"
 	"runtime"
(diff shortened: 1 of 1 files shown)


dragonly:
@cosmtrek Hi! Could you please take a look at this? Thanks!

dragonly:
Any updates?

PR #333, opened 2022-09-29
Update Chinese translation

linweiyuan:


linweiyuan:
No response for a long time, again, so close, please do not re-open.

PR #337, opened 2022-10-17
Add `include_file` option

silverwind:
Useful option to specify individual files to watch. I renamed the `watchDir` function to `watchPath` because it can also handle single files fine (`filepath.Walk` emits a single file if passed a file path).

Fixes: https://github.com/cosmtrek/air/issues/307

Example of it working with a pure `include_dir` and `include_file` config:

image


silverwind:
Will see to adding a test as well.

silverwind:
Good to go now.

xiantang (approved):
LGTM totally

PR #342, opened 2022-10-24
Add the tutorial for docker users

kenilt:
I really love Air features. However, when I started dockerize my Golang app, Air did not work. I started some search and found some alternative libs, but I still love Air. So I found a way to make air work with docker-compose (not work by only `docker run` command with `volume` flag for some reason)

For this PR I just add Dockerfile and docker-compose.yaml examples for Docker users who don't want to use air images with fixed Golang version and Linux distro. I think it will help someone who beginner to Air and Docker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,7 +164,7 @@ services:
 ## Installation and Usage for Docker users who don't want to use air image
 
 `Dockerfile`
-```
+```Dockerfile
 # Choose whatever you want, version >= 1.16
 FROM golang:1.19-alpine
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -161,6 +161,39 @@ services:
 
 `air -d` prints all logs.
 
+## Installation and Usage for Docker users who don't want to use air image
+
+`Dockerfile`
+```
```
xiantang, line 167:
```dockerfile

Comment (xiantang) [reply README.md:167]:
```

`docker-compose.yaml`
```yaml
```


```
`docker-compose.yaml`
```yaml

```

xiantang (approved):

mihir-karbelkar-wiz:
@kenilt thanks for this! Is there a way to run air without doing go install air@latest? I have 5-6 containers where I need to run air, is there a way to not install it for all containers?

PR #343, opened 2022-10-24
change build delay to zero

xiantang:
close https://github.com/cosmtrek/air/issues/325

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,8 +2,6 @@ name: build
 
 on:
   push:
-  pull_request:
-    branches: [ master ]
 
 jobs:
   build:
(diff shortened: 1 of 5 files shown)


silverwind:
with the new `delay=0` default, I see multiple concurrent builds when saving a file once on macOS in Sublime Text:

```
main.go has changed
main.go has changed
building...
building...
main.go has changed
main.go has changed
building...
  >  Building binary...
  >  Building binary...
  >  Building binary...
go build ...
go build ...
go build ...
running...
running...
running...
```

Maybe this is something specific to macOS, but I'm definitely saving the file only once.

silverwind:
delay 0 is also very problematic while switching git branches where every changed file triggers a rebuild, I had to revert it to previous value of 1000 in https://github.com/go-gitea/gitea/pull/24290.

PR #348, opened 2022-10-31
Feat clear on rebuild

mreza0100:
The current clear command fmt.Print("\033[2J") is not completely clear terminal as we can back to it with scroll.
I understand that some people may like to keep the logs and come back with a scroll.
So I added a new param as keep_scroll too.
Also added:
```
Screen: cfgScreen{
	ClearOnRebuild: false,
	KeepScroll:     false,
},
``` 
So people can see the options

### DIFF
--- a/runner/config.go
+++ b/runner/config.go
@@ -240,7 +240,7 @@ func defaultConfig() Config {
 		Misc:        misc,
 		Screen: cfgScreen{
 			ClearOnRebuild: false,
-			KeepScroll:     false,
+			KeepScroll:     true,
 		},
 	}
 }
--- a/runner/config.go
+++ b/runner/config.go
@@ -86,6 +86,7 @@ type cfgMisc struct {
 
 type cfgScreen struct {
 	ClearOnRebuild bool `toml:"clear_on_rebuild"`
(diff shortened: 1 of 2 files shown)


xiantang:
will review this weekend

runner/config.go
```diff
@@ -237,6 +238,10 @@ func defaultConfig() Config {
 		Color:       color,
 		Log:         log,
 		Misc:        misc,
+		Screen: cfgScreen{
+			ClearOnRebuild: false,
+			KeepScroll:     false,
```
silverwind, line 243: I'd argue default should be `true` because a full clear is destructive to terminal history and generally not what people would want.

mreza0100: I see your point.

runner/config.go
```diff
@@ -237,6 +238,10 @@ func defaultConfig() Config {
 		Color:       color,
 		Log:         log,
 		Misc:        misc,
+		Screen: cfgScreen{
+			ClearOnRebuild: false,
+			KeepScroll:     true,
```
mreza0100, line 243: Done

xiantang (approved):

xiantang:
LGTM

PR #358, opened 2022-11-21
Make `include_file` overrule `include_ext`

silverwind:
A file can now be in `include_file` without having its extension in `include_ext`, which I generally think is the expected behaviour because explicit filenames are more specific than extensions.

Fixes: https://github.com/cosmtrek/air/issues/350

### DIFF
--- a/runner/engine.go
+++ b/runner/engine.go
@@ -186,7 +186,7 @@ func (e *Engine) cacheFileChecksums(root string) error {
 			}
 		}
 
-		if e.isExcludeFile(path) || !e.isIncludeExt(path) {
+		if e.isExcludeFile(path) || !e.isIncludeExt(ev.Name) && !e.checkIncludeFile(path) {
 			e.watcherDebug("!exclude checksum %s", e.config.rel(path))
 			return nil
 		}
@@ -251,7 +251,7 @@ func (e *Engine) watchPath(path string) error {
 				if excludeRegex {
 					break
 				}
-				if !e.isIncludeExt(ev.Name) {
+				if !e.isIncludeExt(ev.Name) && !e.checkIncludeFile(path) {
 					break
 				}
 				e.watcherDebug("%s has changed", e.config.rel(ev.Name))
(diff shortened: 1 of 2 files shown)


jamesgoodhouse (approved):

runner/engine.go
```diff
@@ -186,7 +186,7 @@ func (e *Engine) cacheFileChecksums(root string) error {
 			}
 		}
 
-		if e.isExcludeFile(path) || !e.isIncludeExt(path) {
+		if e.isExcludeFile(path) || !e.isIncludeExt(ev.Name) && !e.checkIncludeFile(path) {
```
silverwind, line 189:
```suggestion
		if e.isExcludeFile(path) || !e.isIncludeExt(path) && !e.checkIncludeFile(path) {
```

runner/engine.go
```diff
@@ -251,7 +251,7 @@ func (e *Engine) watchPath(path string) error {
 				if excludeRegex {
 					break
 				}
-				if !e.isIncludeExt(ev.Name) {
+				if !e.isIncludeExt(ev.Name) && !e.checkIncludeFile(path) {
```
silverwind, line 254: I guess the argument here may need to be `ev.Name`, I will re-verify.

zhb127:
Is there anything problem with this pull request?
I see it is marked as draft?

runner/engine.go
zhb127: I test the code below, it's ok:
```
if !e.isIncludeExt(ev.Name) && !e.checkIncludeFile(ev.Name) {
```

I don't know, why marked this pull request as draft, can i create new pull request base on the code.

zhb127: > I guess the argument here may need to be `ev.Name`, I will re-verify.

I test the code below, it's ok:
```
if !e.isIncludeExt(ev.Name) && !e.checkIncludeFile(ev.Name) {
```

I don't know, why marked this pull request as draft, can i recreate new pull request base on the code?

silverwind:
I put it on draft because of https://github.com/cosmtrek/air/pull/358#discussion_r1047117430, but then forgot about the PR.

silverwind:
Your PR does my suggested change so let's move to that one.

PR #363, opened 2022-12-08
WIP: feat: support executing bin again when bin exit

### LINKED ISSUE #353: Air should support executing binary again when executable exit
Currently it doesn't execute binary again when exit. It does only when file changes. `Air` should exit with the same status code as exited binary **OR** rerun binary again with optional configurable delay. Both features should be configurable and may be disabled on default to keep backward compatibility.

**Use case:**
Microservices or long-running applications can be run as services under Docker Compose with the `restart: always` option, Docker Swarm or other container orchestration solution. For development purpose `Air` can be also run in these environments. But some applications/frameworks terminate running application for example because of unreachable message broker or database service instance. And orchestration rerun them again. But with `Air` it never happens. Because `Air` never exit or rerun binary when binary exit by itself.

amikai:
close #353

### DIFF
--- a/runner/engine.go
+++ b/runner/engine.go
@@ -433,15 +433,24 @@ func (e *Engine) building() error {
 func (e *Engine) runBin() error {
 	// control killFunc should be kill or not
 	killCh := make(chan struct{})
+	wg := sync.WaitGroup{}
 	go func() {
 		// listen to binStopCh
 		// cleanup() will close binStopCh when engine stop
 		// start() will close binStopCh when file changed
 		<-e.binStopCh
 		close(killCh)
+
+		select {
+		case <-e.exitCh:
+			wg.Wait()
+			close(e.canExit)
+		default:
+		}
 	}()
 
-	killFunc := func(cmd *exec.Cmd, stdout io.ReadCloser, stderr io.ReadCloser, killCh chan struct{}, processExit chan struct{}) {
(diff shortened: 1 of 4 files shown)


xiantang:
work in progress?

runner/engine.go
```diff
@@ -431,31 +431,37 @@ func (e *Engine) building() error {
 }
```
xiantang, line 432: we need to add a new flag to support this, and disable this behavior by default

amikai: ok

xiantang:
I mean you need add a test case like this: when executing bin exit should restart the bin again

xiantang:
Do we need to append a document for this?

runner/engine_test.go
```diff
@@ -110,6 +111,87 @@ func TestRegexes(t *testing.T) {
...
+	}
+
+	time.Sleep(time.Second * 1)
+	// stop engine
+	engine.Stop()
+	time.Sleep(time.Second * 1)
+	t.Logf("engine stopped")
+
+	roundAfterChange := atomic.LoadUint64(&engine.round)
+	fmt.Println(roundBeforeChange)
```
amikai, line 188: delete this

amikai:
> Do we need to append a document for this?

Finish 1c7171469e52d6b693e72794258c36983ea7892e

PR #369, opened 2023-01-03
Log output doesn't show until close #50

### LINKED ISSUE #50: Log output doesn't show until close
Unsure this is related to my app or air itself but on Windows when I use `log.Print` etc for anything the output doesn't show until I cancel out of the running app and then it will dump everything to stdout.    

Haven't tested another application yet to see if it's me doing something but thought I would just log it here first and close if it's my fault.

iftikharqureshi:
This seems to work on Ubuntu and Windows.

### DIFF
--- a/runner/engine.go
+++ b/runner/engine.go
@@ -439,8 +439,13 @@ func (e *Engine) runBin() error {
 	if err != nil {
 		return err
 	}
+
 	go func() {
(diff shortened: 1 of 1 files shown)


xiantang:
thanks

xiantang (approved):

PR #373, opened 2023-01-11
test: restore the cwd for every test

amikai:
This test of 078d8ffa29e14c4d462c7ab825188895a32abeaf failed, because of some testing function use the `os.Chdir` but not restore it before exit.


xiantang (approved):
LGTM

PR #389, opened 2023-02-02
Update engine.go

lukaswork:
The old version did not print out any log.print or other log outputs under windows (in powershell/cmd/git bash) for us.  After that change it worked. It is just a hacky solution currently. Happily awaiting feedback if this is actually a soultion.

Thanks


anton7r:
Fixes the issue on Windows, thanks! 🙌

Tested it also inside WSL and there seems to be no issues there.

xiantang:
the test cases in linux and mac os is passed, but i cannot test in windows bcs I dont have windows device.

Issue #58, opened 2020-03-25
how to run air with environment variables


shmilyoo:
- this line blew is not work
`cmd = "MONGODB_HOST=127.0.0.1:27017 MONGODB_USER=test MONGODB_PASSWORD=test REDIS_ADDR=127.0.0.1 GIN_MODE=debug go run *.go"`

- this line blew works well
`MONGODB_HOST=127.0.0.1:27017 MONGODB_USER=test MONGODB_PASSWORD=test REDIS_ADDR=127.0.0.1 GIN_MODE=debug air`

### .air.conf
```
# Config file for [Air](https://github.com/cosmtrek/air) in TOML format

# Working directory
# . or absolute path, please note that the directories following must be under root.
root = "." 
tmp_dir = "tmp"

[runner]
# Just plain old shell command. You could use `make` as well.
cmd = "MONGODB_HOST=127.0.0.1:27017 MONGODB_USER=test MONGODB_PASSWORD=test REDIS_ADDR=127.0.0.1 GIN_MODE=debug go run *.go"
# Binary file yields from `cmd`.
bin = "tmp/main"
# Customize binary.
# full_bin = "APP_ENV=dev APP_USER=air ./tmp/main"
# Watch these filename extensions.
include_ext = ["go", "tpl", "tmpl", "html"]
# Ignore these filename extensions or directories.
exclude_dir = ["assets", "vendor", "frontend/node_modules"]
# Watch these directories if you specified.
include_dir = []
# Exclude files.
exclude_file = []
# It's not necessary to trigger build each time file changes if it's too frequent.
delay = 1000 # ms
# Stop to run old binary when build errors occur.
stop_on_error = true
# This log file places in your tmp_dir.
log = "air_errors.log"

[log]
# Show log time
time = false

[color]
# Customize each part's color. If no color found, use the raw app log.
main = "magenta"
watcher = "cyan"
build = "yellow"
runner = "green"

[misc]
# Delete tmp directory on exit
clean_on_exit = true
```

cosmtrek:
Please use `full_bin` instead of `cmd`, like:

```
full_bin = "MONGODB_HOST=127.0.0.1:27017 MONGODB_USER=test MONGODB_PASSWORD=test REDIS_ADDR=127.0.0.1 GIN_MODE=debug go run *.go"
```

shmilyoo:
> Please use `full_bin` instead of `cmd`, like:
> 
> ```
[...]

PR #405, opened 2023-04-03
Load environment variables from file

ZaninAndrea:
This pull request adds a new key `env_file` to the configuration, the key specifies a file from which the environment variables for running `bin` should be sourced.  
If `env_file` is set, the environment file is sourced every time the `bin` is run, so that it works as expected when "env" is in `include_ext` and the env file is updated.

The new feature was suggested in [#47](https://github.com/cosmtrek/air/pull/47#issuecomment-652642051) and removes the need for less elegant solutions like the one suggested in [#58](https://github.com/cosmtrek/air/issues/58#issuecomment-898537034).

### DIFF
--- a/air_example.toml
+++ b/air_example.toml
@@ -44,6 +44,8 @@ rerun = false
 rerun_delay = 500
 # Add additional arguments when running binary (bin/full_bin). Will run './tmp/main hello world'.
 args_bin = ["hello", "world"]
+# Load environment variables from a file
+env_file = ".env"
 
 [log]
 # Show log time
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ require (
 
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/joho/godotenv v1.5.1 // indirect
 	github.com/mattn/go-colorable v0.1.13 // indirect
 	github.com/mattn/go-isatty v0.0.17 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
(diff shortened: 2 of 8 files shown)


silverwind:
Does it also work when `cmd` and `full_bin` run? Please add some docs on where it exactly applies.

mfechner:
I think using `cmd` or `full_bin` will not work if you use air on e.g. Linux and Windows for the same project, as both environment are defining env variables differently. I think using a `env_file` configuration parameter is the only way to make this feature for all operating systems available.

ZaninAndrea:
The implementation doesn't break existing setups, if the `env_file` parameter is not specified it defaults to `""` (due to how the parameters are parsed) and thus no environment file is loaded. I added the `env_file = ".env"` parameter in the example configuration as a form of documentation, but that file is meant to be modified by the user when doing the first configuration.

The environment file is loaded when running `bin`/`full_bin` but not when running `cmd`. I think that this feature is much more important for `bin`/`full_bin` rather than `cmd`, because the environment variables used in the build command can generally be safely committed in git, while the variables used at runtime may contain secrets; either way we could extend this implementation with another `cmd_env_file` (or something similar) quite easily.

Multiple environment files are indeed not supported, I could extend this to accept a list of files instead of a single file.

whoiscarlo:
Any updates on this?

whoiscarlo:
> I think using `cmd` or `full_bin` will not work if you use air on e.g. Linux and Windows for the same project, as both environment are defining env variables differently. I think using a `env_file` configuration parameter is the only way to make this feature for all operating systems available.

I found that it also doesn't work on Mac Os, not sure what's missing but my environment variables aren't being brought over =/

PR #409, opened 2023-04-12
Add support for polling for file changes

### LINKED ISSUE #360: Hot reload does not work with Virtualbox shared folders
FYI Fsnotify is unable to detect file changes in Virtualbox shared folders(which is my use case, but this probably affects a variety of fs setups) so hot reload would not work

See relevant discussion https://github.com/fsnotify/fsnotify/issues/152
Talks of adding a polling fallback https://github.com/fsnotify/fsnotify/issues/9
Webpack/Nodemon solved this by polling too, via an option ie `poll = 1000ms`

Supposedly working(?) polling implementation from Moby based on Fsnotify
https://github.com/moby/moby/commit/c609523a8c006fdd404a8b99be2567401c3ec1b9
used in Hugo https://github.com/gohugoio/hugo/tree/master/watcher/filenotify

The polling issue in Fsnotify is 8 years old though. Makes sense to use the Moby fix in the meantime?

svanharmelen:
I thought about just copying the files and adding a notice in the header of each copied file, but in the end decided to just use the modules instead. Let me know if you prefer otherwise...

And of course let me know what you think of this PR 😉 It fixes #360 and is needed for our own dev setup, so I hope this can be merged so we can keep using the upstream code instead of our fork.

Thanks!


cosmtrek:
Thanks a lot~

PR #411, opened 2023-04-13
Prevent double delays and add some docs

svanharmelen:
I just noticed that reusing `delay` for the poll interval will cause an unintended double delay (first the delay is used as interval, and when a change happened delay is used to wait before building after the change).

Adding a dedicated `poll_interval` parameter allows people to prevent that and apply a more fine-grained configuration.

Also added the new params to the example toml file with some docs which I forgot to do in the initial PR and "fixed" a few linter warnings (if wanted I can of course remove that commit from the PR!).

Thanks again!

### DIFF
--- a/runner/engine_test.go
+++ b/runner/engine_test.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"log"
 	"net"
 	"os"
 	"os/exec"
@@ -508,10 +509,7 @@ func checkPortConnectionRefused(port int) bool {
 			_ = conn.Close()
 		}
 	}()
-	if errors.Is(err, syscall.ECONNREFUSED) {
-		return true
-	}
-	return false
+	return errors.Is(err, syscall.ECONNREFUSED)
 }
 
 func checkPortHaveBeenUsed(port int) bool {
@@ -572,6 +570,9 @@ func main() {
 		return err
 	}
 	_, err = file.WriteString(code)
+	if err != nil {
+		return err
+	}
 
 	// generate go mod file
 	mod := `module air.sample.com
@@ -604,6 +605,9 @@ func main() {
 		return err
 	}
 	_, err = file.WriteString(code)
+	if err != nil {
+		return err
+	}
 
 	// generate go mod file
 	mod := `module air.sample.com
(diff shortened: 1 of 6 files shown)


runner/engine_test.go
```diff
@@ -895,11 +902,11 @@ include_ext = ["sh"]
 include_dir = ["nonexist"] # prevent default "." watch from taking effect
 include_file = ["main.sh"]
 `
-	if err := ioutil.WriteFile(dftTOML, []byte(config), 0644); err != nil {
+	if err := ioutil.WriteFile(dftTOML, []byte(config), 0o644); err != nil {
```
cosmtrek, line 905: Is there a need to change?

runner/engine_test.go
```diff
@@ -688,12 +695,12 @@ func TestRebuildWhenRunCmdUsingDLV(t *testing.T) {
 	go func() {
 		file, err := os.OpenFile("main.go", os.O_APPEND|os.O_WRONLY, 0o644)
 		if err != nil {
-			t.Fatalf("Should not be fail: %s.", err)
+			log.Fatalf("Should not be fail: %s.", err)
```
cosmtrek, line 698: What are the benefits of changing it?

cosmtrek:
Thanks for the contribution. I've reviewed the code and left some comments.

runner/engine_test.go
svanharmelen: To be really honest, I don't know... But my LSP (Gopls) gave me this error and warning:

image

image

But no problem to revert this, just let me know!

runner/engine_test.go
svanharmelen: Again, just let me know if you prefer the PR without this...

cosmtrek: OK

runner/engine_test.go
cosmtrek: OK

svanharmelen:
Thanks! If possible, could you maybe set a new tag so we can use that instead of `@latest`?

svanharmelen:
Just noticed the tag, thanks for being so responsive and helpful 👍🏻 Much appreciated!

silverwind:
Sounds this may fix what I've seen in https://github.com/cosmtrek/air/pull/343#issuecomment-1305897741.

PR #418, opened 2023-04-27
Make air work with Colima

zachgrayio:
Air probably shouldn't filter out chmod events, as they're valid filewatcher events dispatched by popular tools specifically for livereload.

This patch fixes Air when using Colima instead of Docker Desktop for Mac, as most users are moving away from Docker Desktop in favor of Colima after the licensing changes.

Related: https://github.com/abiosoft/colima/issues/261#issuecomment-1484677394

### DIFF
--- a/runner/engine.go
+++ b/runner/engine.go
@@ -239,9 +239,6 @@ func (e *Engine) watchPath(path string) error {
 				return
 			case ev := <-e.watcher.Events():
 				e.mainDebug("event: %+v", ev)
-				if !validEvent(ev) {
-					break
-				}
 				if isDir(ev.Name) {
 					e.watchNewDir(ev.Name, removeEvent(ev))
 					break
--- a/runner/util.go
+++ b/runner/util.go
@@ -215,12 +215,6 @@ func isDir(path string) bool {
 	return i.IsDir()
 }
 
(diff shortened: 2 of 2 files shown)


zachgrayio:
Verified this commit works as expected on M1 MBP

```
sysctl -n machdep.cpu.brand_string
Apple M1 Pro
```

```
sw_vers
ProductName:		macOS
ProductVersion:		13.2.1
BuildVersion:		22D68
```
and 
- colima version `HEAD-9575e08` started with `colima start --mount-inotify`
- docker `version 23.0.4, build f480fb1e37`
- docker-compose `version 2.17.3`

cosmtrek:
@zachgrayio Thanks~

runner/engine.go
```diff
@@ -239,9 +239,6 @@ func (e *Engine) watchPath(path string) error {
 				return
 			case ev := <-e.watcher.Events():
 				e.mainDebug("event: %+v", ev)
-				if !validEvent(ev) {
```
cosmtrek, line 242: If the check is removed, the file checksum will be broken. I need to keep this logic...

mvhatch: what is the consequence of adding `CHMOD` to the list of valid events?

PR #433, opened 2023-06-16
fix air version

gaols:
fix: air version


xiantang:
thank you, but we use buildinfo to fix this

PR #449, opened 2023-08-09
chore: bump Go to 1.21

ethancdaniel:
Previous PR for bumping to 1.20 for reference: https://github.com/cosmtrek/air/pull/394.


ethancdaniel:
Thanks for merging, could we get a new release? @xiantang

PR #450, opened 2023-08-17
build binaries with CGO_ENABLED=0

mosaxiv:
* CGO_ENABLED should be set to "0" when application running in alpine linux or it will be failed to start.


xiantang:
good catch

Issue #341, opened 2022-10-23
support for yaml configuration file?


dirsigler:
It sounds quite trivial, but I would be interested if there is support to also write the air configuration file in yaml.

If not, I would love to see support for a yaml configuration file.
All of my configuration files in a project are already either JSON or YAML files which means I also have the linting and extensions already in place for these. TOML is quite rare.

xiantang:
u can support YAML file, PR is welcome

xiantang:
if u want, u can also support JSON. haha

xiantang:
I will close this Merge request, if PR  is ready, plz tag me

christopher-kleine:
@xiantang Could you review PR #455 ? It shouldn't break anything and is backwards compatible.
Let me know if there's something missing.

christopher-kleine:
> u can support YAML file, PR is welcome

Is this still valid? My PR for exactly is open for a few weeks now.
Not even a quick response.
And I'm not the only one who is waiting for a response.

Is this project still maintained? Except for like 2 PRs, the whole interaction seems a bit lacking.

christopher-kleine:
As mentioned, I closed my PR and start managing my own fork now.

PR #455, opened 2023-08-21
feat: Add support for YAML files

christopher-kleine:
This addresses #341

I tested this with an existing config. It's also compatible with existing configs.
But since I didn't invest too much into this, let me know if I forgot anything.

Another change: At start it shows the used config file (if any).

TODO:
- [x] Support for `.air.yml` and `.air.yaml` (Default: `.air.toml`)
- [x] `air init toml`, `air init yaml` and `air init yml`

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,12 @@ You can initialize the `.air.toml` configuration file to the current directory w
 air init
 ```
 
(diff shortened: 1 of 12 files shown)


christopher-kleine:
I take my PR back. I guess I'll manage my own fork now. Maybe this way some changes get actually done.

PR #457, opened 2023-08-27
fix: Prevent duplicate processes when multiple buildRuns run simultaneously

### LINKED ISSUE #404: run multiple cmd one live reload
I use quicktample on my web, but it also leave me are difficult on using air, 

can i use air auto build my qtpl to qtpl.go by cma qtc then run go run to run the code??

### LINKED ISSUE #421: address already in use
System Version: macOS 13.3.1 (22E261)
      Kernel Version: Darwin 22.4.0

When I used air, except for the first time, hot loads always had port occupancy issues
image
This is the port usage using Chrome
image

### LINKED ISSUE #426: Duplicate process after reload
image

image
Code changes under ./cmd are not detected and not reloading

and cd to `./cmd` and run `air` with config under `./cmd`  then it will work, tried change `root = "."` to `root = "./cmd"` but still not detecting changes

image

toga4:
Hi maintainers, this PR fixes the behavior that air duplicates processses when multiple buildRuns run simultaneously.
Closes #404, Closes #421, Closes #426.

To reproduce issue:

main.go
```go
package main

import (
	"log"
	"net/http"
	"os"
)

func main() {
	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		w.Write([]byte("Hello World!"))
	})

	log.Print(os.Getpid(), " Starting server...")
	if err := http.ListenAndServe(":8080", nil); err != nil {
		log.Fatal(os.Getpid(), err)
	}
}
```

Run `air -d` and save a file multiple times in a short period of time, then air executes multiple processes.
air

This behavior appears from v1.41.0, and seems to be brought by the following pull request.
- #363 

I looked into what was making the difference, and it seems that the following code that was in place before the change was important.
https://github.com/cosmtrek/air/blob/d0a697adcf852559ab6fff59aa164b18f62571a0/runner/engine.go#L486-L490

This PR revive that code, and makes
- killFunc is always resumed by binStopCh close.
- killCh is for break rerun loop only.
- handle exitCh after rerun loop breaks.

### DIFF
--- a/runner/engine.go
+++ b/runner/engine.go
@@ -435,10 +435,6 @@ func (e *Engine) runBin() error {
 	killFunc := func(cmd *exec.Cmd, stdout io.ReadCloser, stderr io.ReadCloser, killCh chan struct{}, processExit chan struct{}, wg *sync.WaitGroup) {
 		defer wg.Done()
 		select {
-		// the process haven't exited yet, kill it
-		case <-killCh:
-			break
-
 		// listen to binStopCh
 		// cleanup() will close binStopCh when engine stop
 		// start() will close binStopCh when file changed
--- a/runner/engine.go
+++ b/runner/engine.go
@@ -432,29 +432,18 @@ func (e *Engine) building() error {
 }
 
 func (e *Engine) runBin() error {
-	// control killFunc should be kill or not
-	killCh := make(chan struct{})
-	wg := sync.WaitGroup{}
-	go func() {
-		// listen to binStopCh
-		// cleanup() will close binStopCh when engine stop
-		// start() will close binStopCh when file changed
-		<-e.binStopCh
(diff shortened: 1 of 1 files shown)


toga4:
Confirmed this change works as expected (on M1 Mac).
image

toga4:
Hi, @cosmtrek @xiantang 
Is there any chance you could review this PR? I'd appreciate your feedback.

suntong:
Hope the PR can be merged soon @cosmtrek @xiantang. thx!

xiantang (approved):
LGTM totally

xiantang:
I know the root cause now, it's because when you save your code in a short period of time, air will start duplicate processes, but doesn't have any lock for it.
thank you good catch bug

Issue #216, opened 2021-11-25
port address already in use


guojunhao118:
![WX20211125-105752](https://user-images.githubusercontent.com/21166467/143372710-d67c1e31-0ce1-448d-b63d-e22e9a826563.png)
I get this problem in version 1.27.5,and i use mac,how can i to solve this?

xiantang:
could u provide the exmaple that i can reproduce

xiantang:
you can try 1.27.6

marekmartins:
I have exactly the same problem using it under WSL2 in Win11. Problem started with version 1.27.5. Also 1.27.6 is affected aswell.

For reproducing dummy HTTP server:
`
package main

import (
	"log"
	"net/http"
)

func main() {
	log.Fatal(http.ListenAndServe(":8080", nil))
}
`
![image](https://user-images.githubusercontent.com/55179666/143444181-d32ab6c8-8a38-4cd4-9af7-0d2e5292fba0.png)

xiantang:
could u use air -d and provide log？

marekmartins ***@***.***>于2021年11月25日 周四20:45写道：

marekmartins:
Sure
```
[crypter] event: "/app/main.go": WRITE
[crypter] event: "/app/main.go": WRITE
[crypter] main.go has changed
[crypter] main.go has changed
[crypter] flushing events
[crypter] main.go has changed
[crypter] building...
[crypter] trying to kill pid 119, cmd [/bin/sh -c /app/tmp/main]
[crypter] running...
[crypter] running process pid 229
[crypter] 2021/11/26 03:35:44 listen tcp :8080: bind: address already in use
```

xiantang:
I have a windows device, i will reproduce this issue, thanks! @marekmartins

xiantang:
which subsystem did u use. I have use ubuntu to try to reproduce this issue but works good. @marekmartins

xiantang:
![1637941783(1)](https://user-images.githubusercontent.com/34479567/143605521-643a3fc8-ee6c-4115-ad0d-755e550b72d2.png)
windows
![image](https://user-images.githubusercontent.com/34479567/143605675-fccd6792-38cd-4986-aa6d-c1a2431ffd3a.png)
ubuntu

marekmartins:
Ubuntu, are You using WSL2? Version 2 is important :)

In powershell:
[...]

PR #462, opened 2023-08-30
Interpret value of kill_delay as milliseconds

lseppala:
`kill_delay` is parsed as a Duration ([ref](https://github.com/cosmtrek/air/blob/af962ce3f6f1a048642d5c70f97aa525792b6489/runner/config.go#L55)). When given as an integer (as in the [example config](https://github.om/cosmtrek/air/blob/af962ce3f6f1a048642d5c70f97aa525792b6489/air_example.toml#L44)), it is interpreted as nanoseconds, not milliseconds. The sleep will be too short to allow a program to respond to the interrupt and stop cleanly. This change allows the number to be interpreted as milliseconds.

This change is made to preserve backwards compatibility. If the value is specified as an value under 1 millisecond/1 000 000 nanoseconds, like `500` in the example config, the value will be interpreted as the number of milliseconds. For users who figured out that you may specify the time as a duration string like `”2s”` or use large nanosecond integers like `2000000000` for 2 seconds, the behavior will remain the same.

An alternative is changing `kill_delay` to be parsed as an integer, like `delay` and `rerun_delay`. This will break users who used duration strings or large numbers.

This may be the root cause of the problem reported in #216, as a sufficient value for `kill_delay` to allow a program to stop gracefully solved this problem for me.

### DIFF
--- a/runner/config.go
+++ b/runner/config.go
@@ -340,6 +340,17 @@ func (c *Config) rerunDelay() time.Duration {
 	return time.Duration(c.Build.RerunDelay) * time.Millisecond
 }
 
+func (c *Config) killDelay() time.Duration {
+	// kill_delay can be specified as an integer or duration string
+	// interpret as milliseconds if less than the value of 1 millisecond
+	if c.Build.KillDelay < time.Millisecond {
+		return c.Build.KillDelay * time.Millisecond
+	} else {
+		// normalize kill delay to milliseconds
+		return time.Duration(c.Build.KillDelay.Milliseconds()) * time.Millisecond
+	}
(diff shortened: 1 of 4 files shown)


satetsu888:
related link https://github.com/cosmtrek/air/pull/397

But I think this solution is better, with nice backward compatibility considerations.

xiantang:
I think the code is great but could u add more test cases for it, once the code is covered i will merge it

lseppala:
@xiantang Tests cases added. Let me know if you want something different.

xiantang (approved):

PR #474, opened 2023-10-16
fallback to buildinfo when airversion and goversion is empty

xiantang:
close https://github.com/cosmtrek/air/issues/434


xiantang:
need to fix test, but not related with this PR

PR #489, opened 2023-10-20
Add global default config

sarjann:
Added the ability to specify your own default config.
Think it's a nice convenience, in my case, I wanted to change the tmp directory to a hidden directory for example
and it's a bit annoying to have to update the air.toml in each project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,7 @@ air -c .air.toml
 ```
 
 You can initialize the `.air.toml` configuration file to the current directory with the default settings running the following command.
+You can also create a global default for `.air.toml` in `~/.air.toml`. 
 
 ```bash
 air init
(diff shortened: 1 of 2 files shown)


xiantang:
seems good but I i want to know is this the common case ?

sarjann:
> seems good but I i want to know is this the common case ?

Hard to say without asking other people, but this doesn't break backwards compatibility and I think it's a nice feature. Also I think a lot of people will probably have their own defaults that they like.

sarjann:
Not sure if there are any additional changes you want to make or if you're happy for this to be merged. @xiantang 
If you're fine with the changes I can add some tests (lines aren't currently covered), I'm having some trouble running tests locally but can figure it out.

PR #494, opened 2023-11-04
Add skip_run so build only works on windows

joeyak:
Since there is no `/usr/bin/true` for windows or an executable close to it I added a flag to skip the run step. I originally followed https://github.com/cosmtrek/air/issues/365 by creating an empty go program, but then realized I would have to add `true.exe` to my repo.


praxxus11 (approved):

zoop-btc:
I'm in need of this too since I use nixos but other devs don't the path to "true" won't be the same.

joeyak:
@praxxus11 I've fixed the merge conflicts

xiantang:
```
[build]
  cmd = "go build -o ./tmp/main ."
  bin = "echo"  # Will work on both Windows and Unix-based systems
```
can u try this?

joeyak:
It doesn't look clean, but it does work in the end
image

PR #512, opened 2024-01-05
Add live proxy feature

ndajr:
## Description
This PR implements what is discussed in this issue: https://github.com/cosmtrek/air/issues/141. I started a new webapp using Go templates and would like to have a live reloading functionality (not just hot reloading when files change). The two possible implementations are with websockets or with server-sent events. I chose the latter because it adds the least amount of complexity and no external dependency is required. I'd like some feedback on the config names (port and app_port, or frontend_port and backend_port?) and whether proxy should be enabled by default or not.

## How to use
It is quite simple to use, add the proxy config, in my case I added my views containing the go templates on `include_dir`, and access the proxy port, e.g. `http://localhost:8090`.

## How it works
There are two handlers on the proxy, the root handler `/` which is a catch-all and should proxy all requests and the `/internal/reload` endpoint which implements server-sent events. The root handler proxies all the requests and injects a script that listens to `/internal/reload` and reloads the browser every time air watches a file change.

### DIFF
--- a/air_example.toml
+++ b/air_example.toml
@@ -73,3 +73,9 @@ clean_on_exit = true
 [screen]
 clear_on_rebuild = true
 keep_scroll = true
+
+# Enable live-reloading on the browser.
+[proxy]
+  enabled = true
+  proxy_port = 8090
+  app_port = 8080
--- a/runner/config.go
+++ b/runner/config.go
@@ -31,6 +31,7 @@ type Config struct {
 	Log         cfgLog    `toml:"log"`
 	Misc        cfgMisc   `toml:"misc"`
 	Screen      cfgScreen `toml:"screen"`
+	Proxy       cfgProxy  `toml:"proxy"`
 }
 
 type cfgBuild struct {
@@ -94,6 +95,12 @@ type cfgMisc struct {
 type cfgScreen struct {
 	ClearOnRebuild bool `toml:"clear_on_rebuild"`
 	KeepScroll     bool `toml:"keep_scroll"`
+}
+
+type cfgProxy struct {
+	Enabled   bool `toml:"enabled"`
+	ProxyPort int  `toml:"proxy_port"`
+	AppPort   int  `toml:"app_port"`
 }
 
 type sliceTransformer struct{}
@@ -350,10 +357,9 @@ func (c *Config) killDelay() time.Duration {
 	// interpret as milliseconds if less than the value of 1 millisecond
 	if c.Build.KillDelay < time.Millisecond {
 		return c.Build.KillDelay * time.Millisecond
-	} else {
-		// normalize kill delay to milliseconds
-		return time.Duration(c.Build.KillDelay.Milliseconds()) * time.Millisecond
 	}
+	// normalize kill delay to milliseconds
+	return time.Duration(c.Build.KillDelay.Milliseconds()) * time.Millisecond
 }
 
 func (c *Config) binPath() string {
--- a/runner/engine.go
+++ b/runner/engine.go
@@ -18,6 +18,7 @@ import (
 // Engine ...
 type Engine struct {
 	config    *Config
+	proxy     *Proxy
 	logger    *logger
 	watcher   filenotify.FileWatcher
 	debugMode bool
@@ -48,6 +49,7 @@ func NewEngineWithConfig(cfg *Config, debugMode bool) (*Engine, error) {
 	}
 	e := Engine{
 		config:         cfg,
+		proxy:          NewProxy(&cfg.Proxy),
 		logger:         logger,
 		watcher:        watcher,
 		debugMode:      debugMode,
@@ -310,6 +312,11 @@ func (e *Engine) isModified(filename string) bool {
 
 // Endless loop and never return
 func (e *Engine) start() {
+	if e.config.Proxy.Enabled {
+		go e.proxy.Run()
+		e.mainLog("Proxy server listening on http://localhost%s", e.proxy.server.Addr)
+	}
+
 	e.running = true
 	firstRunCh := make(chan bool, 1)
 	firstRunCh <- true
@@ -535,6 +542,9 @@ func (e *Engine) runBin() error {
 				cmd, stdout, stderr, _ := e.startCmd(command)
 				processExit := make(chan struct{})
 				e.mainDebug("running process pid %v", cmd.Process.Pid)
+				if e.config.Proxy.Enabled {
+					e.proxy.Reload()
+				}
 
 				wg.Add(1)
 				atomic.AddUint64(&e.round, 1)
@@ -578,6 +588,11 @@ func (e *Engine) runBin() error {
 func (e *Engine) cleanup() {
 	e.mainLog("cleaning...")
 	defer e.mainLog("see you again~")
+
+	if e.config.Proxy.Enabled {
+		e.mainDebug("powering down the proxy...")
+		e.proxy.Stop()
+	}
 
 	e.withLock(func() {
 		close(e.binStopCh)
--- a/runner/engine_test.go
+++ b/runner/engine_test.go
@@ -927,6 +927,9 @@ func Test(t *testing.T) {
 	t.Log("testing")
 }
 `)
+	if err != nil {
+		t.Fatal(err)
+	}
 	// run sed
 	// check the file is exist
 	if _, err := os.Stat(dftTOML); err != nil {
--- a/runner/proxy.go
+++ b/runner/proxy.go
@@ -59,12 +59,14 @@ func (p *Proxy) injectLiveReload(origURL string, respBody io.ReadCloser) string 
 	if _, err := buf.ReadFrom(respBody); err != nil {
 		log.Fatalf("failed to convert request body to bytes buffer, err: %+v\n", err)
 	}
-	s := buf.String()
+	original := buf.String()
 
-	body := strings.LastIndex(s, "</body>")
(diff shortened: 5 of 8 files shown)


scottmckendry:
@ndajr I'm trying to test this out in a dockerized chi app. Could I please have some clarification on how this should work?

I've updated my dockerfile to point to your branch and install air as per below:

```dockerfile
FROM golang:1.21

WORKDIR /jacc
COPY . /jacc/
RUN apt-get update && apt-get install wkhtmltopdf -y
RUN go mod download
RUN go mod verify
RUN go install github.com/a-h/templ/cmd/templ@latest
RUN git clone https://github.com/ndajr/air ~/.air && cd ~/.air && go install . && cd -
CMD air -c .air.toml
```

and I've updated my `.air.toml` file with the following:
```toml
[proxy]
  enabled = true
  proxy_port = 8080
  app_port = 3000
```

I've opened up both ports in my docker-compose. My chi app listens on port 3000. Tested live reloading on port 3000 - updated a file, watched air recompile and restart the server but no refresh on the browser side. The changes only appear after a manual refresh. Port 8080 is unresponsive.

Any idea what I could be doing wrong?

ndajr:
Hey @scottmckendry, I have changed the proxy config to accept urls not just ports, so that it works also with docker compose. Please try this: https://github.com/cosmtrek/air/pull/512/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R214

scottmckendry:
Thanks @ndajr but I'm not convinced that was the issue. I'm running the container locally so using the default `http://localhost` URL was probably fine. 

Either way, I've updated the .air.toml config as per your documentation. But still no difference, unfortunately. Perhaps I need some clarity on expected behaviour. If my app is listening on port 3000 and the proxy is configured to port 8080, which URL should I be using to access the app?

What should I be expecting to see after air reloads?

ndajr:
@scottmckendry you should access the proxy url, in your case it's 8080. Proxy will try to call your chi app on port 3000 and you should see the exact same page with an added script before the body html tag, which does the live reloading without you having to change the code of your app. You can try to access it outside of docker and see if it works, you can post here the full docker-compose and Dockerfiles and I can try to reproduce

scottmckendry:
Hi @ndajr 👋

I've made some progress on my end. Turns out I was cloning the wrong branch in my dockerfile. Adding `-b feat-live-proxy` resolved that. Now I'm actually getting the log entry for the proxy server starting up.

Sadly, I'm still getting no response on the proxy URL. I suspect this is due to my docker config. I've pushed my current setup to the [air-proxy-reload](https://github.com/scottmckendry/jacc/tree/air-proxy-reload) branch in the repo if you want to take a look.

Appreciate your help 🙂

ndajr:
Hi @scottmckendry, I think now I've got more clarity on what the issue was. Althougth listening on `localhost:<port>` is more convenient locally and avoids the message "Do you want the application “air” to accept incoming network connections?", it doesn't work with docker, it has to listen on `:<port>`. Sorry for the back and forth, please revert the config to this:

```toml
[proxy]
  enabled = true
  proxy_port = 8080
  app_port = 3000
```

I had to deal with a lot more edge cases parsing the full url and it wasn't worth it. My branch is updated, let me know how it goes.

scottmckendry:
Brilliant! That's working perfectly now. Great stuff.
Hope this gets merged soon, it's a really neat feature👍🏻

enzshew (approved):

zolrath:
When using htmx to perform an update with partials the proxy fails as the html replacement fails due to the lack of body tag.


Clicking this button:
```html
<button hx-get="/user/more" hx-target="#data">
   Load fresh data
</button>

<div id="data"></div>
```

Calls an endpoint which returns this partial:
```golang
package user

templ More() {
  <p>Injected partial</p>
}
```

Which causes this error:
```
{"time":"2024-01-26T17:58:57.853555007-08:00","id":"","remote_ip":"[::1]:41856","host":"localhost:3000","method":"GET","uri":"/user/more","user_agent":"Go-http-client/1.1","status":200,"error":"","latency":5751,"latency_human":"5.751µs","bytes_in":0,"bytes_out":0}
2024/01/26 17:58:57 invalid html page, missing the body tag
```

ndajr:
Hi @zolrath, thanks for testing the feature with htmx! Initially I was locating the end of the body tag just to know where to insert the live reload script. As we don't want each partial template to be listening to live reload events, just the main page, I pushed a change to proxy the original fragment in case the body tag is missing (instead of erroring), so only the main page will do live reloading.

zolrath:
Hm, my POST doesn't seem to work with the latest version of the proxy.

When I submit a form with an hx-post to an endpoint with `go run` the results are:
```json
{"time":"2024-01-28T23:45:29.537573732-08:00","id":"","remote_ip":"::1","host":"localhost:3000","method":"POST","uri":"/win","user_agent":"Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/121.0.0.0 Safari/537.36 Edg/121.0.0.0","status":200,"error":"","latency":46378798,"latency_human":"46.378798ms","bytes_in":15,"bytes_out":0}
```

If I am running it through the air proxy:
```json
{"time":"2024-01-28T23:46:26.249693344-08:00","id":"","remote_ip":"[::1]:45426","host":"localhost:3000","method":"POST","uri":"/win","user_agent":"Go-http-client/1.1","status":200,"error":"","latency":9799,"latency_human":"9.799µs","bytes_in":0,"bytes_out":0}
```

ndajr:
Hi @zolrath, sorry for the delay, this should be fixed by now. Let me know how it goes

zolrath:
It seems to post properly now!
Though, the proxy live reload seems to be appending the original url to the end of the proxy url instead of simply the path.

image

ndajr:
@zolrath, good finding, I pushed yet another fix, this time actually simplifying the reload script to simply refresh the page which should be more reliable. On the other hand, you will need some `hx-get` with `hx-trigger="load"` on your dynamic content, before after refresh the previous DOM changes will be gone. I created myself an htmx small project and it works fine, I hope it works for you as well.

evanlurvey:
@ndajr cool feature! Are you familiar with the [httputil](https://pkg.go.dev/net/http/httputil#NewSingleHostReverseProxy) package? There is a reverse proxy that does a lot of the heavy lifting for you.

zoop-btc:
I tried out this branch and it works, but since I don't want to compromise on my port I put the refresh script in manually.
It runs into a cors issue that way. Can you fix this?
Doesn't happen in templs proxy btw, which does the same thing as this.

Dieman89:
Is there any update on this?

ndajr:
I'm happy to move this feature forward, do adjustments or fixes (like CORS or httputil mentioned above), but first I'd like to hear from @cosmtrek or one of the maintainers if they want this

popisdead:
This is the killer feature. Imagine having hot reload without live reload and press f5 for every tailwind tweak...

jkeddari:
i use this branch, and it's work perfectly to refresh go-htmx-templ project with tailwindcss rebuild.
It's a game changer !

avilesj:
This is amazing work @ndajr

xiantang:
well done! I will review this PR within next week.

xiantang (approved):

xiantang:
LGTM. WEEEEELL done!!!

xiantang:
I learnt a lot from your PR. tks

conrad760:
Great PR

scottmckendry:
> Thanks for the push, but how can I make this work in container?

You need to open both your app and proxy ports on the container. Your container should also have air as an entry point.

If you want an example you can take a look at the dev.dockerfile and docker-compose configs in one of my projects here:

https://github.com/scottmckendry/mnemstart

DevDorrejo:
> > Thanks for the push, but how can I make this work in container?
> 
> You need to open both your app and proxy ports on the container. Your container should also have air as an entry point.
> 
> If you want an example you can take a look at the dev.dockerfile and docker-compose configs in one of my projects here:
> 
> https://github.com/scottmckendry/mnemstart

I am totally new in Golang.

I didn't explain myself correctly, I want to the website refresh when updating a tmpl/html, or statics file, is this possible and how?

this is my air.toml:

```
root = "."
testdata_dir = "testdata"
tmp_dir = "tmp"

[build]
  args_bin = []
  #bin = "./tmp/main"
  #cmd = "go build -o ./tmp/main ."
  #full_bin = ""
  bin = ""
  cmd = "echo 'Skipping build step in Air...'"
  full_bin = "go run ./cmd/web"
  delay = 1000
  exclude_dir = ["assets", "tmp", "vendor", "testdata"]
  exclude_file = []
  exclude_regex = ["_test.go"]
  exclude_unchanged = false
  follow_symlink = false
  include_dir = []
  include_file = []
  kill_delay = "0s"
  log = "build-errors.log"
  poll = true
  poll_interval = 0
  post_cmd = []
  pre_cmd = []
  rerun = false
  rerun_delay = 500
  send_interrupt = true
  stop_on_error = true

[color]
  app = ""
  build = "yellow"
  main = "magenta"
  runner = "green"
  watcher = "cyan"

[log]
  main_only = false
  silent = false
  time = false

[misc]
  clean_on_exit = true

[proxy]
  enabled = true
  proxy_port = 4001
  app_port = 4000

[screen]
  clear_on_rebuild = false
  keep_scroll = false
```
I am executing air from container, but go project is in host and not container.

Tree:
image

PR #515, opened 2024-01-19
Fix garbled encoding issue when build fails on non-Latin Windows systems

fanyang89:
The default encoding in Go is UTF-8. However, when `cmd.exe` is run on non-Latin Windows systems, it may use an encoding other than UTF-8, which can lead to garbled text.

This patch addresses the issue by running `chcp` to set the code page to UTF-8 before executing commands.

Reproduce(run on a non-latin Windows):

```bash
air init
```

Then modify the `.air.toml`:

```
  bin = "tmp\\main.exe"
  cmd = "go build -o ./tmp/main.exe ."
```

To(leave `tmp_dir` alone)

```
  bin = "bin\\main.exe"
  cmd = "go build -o ./bin/main.exe ."
```

Before this patch:

image

After:

image


kondr1:
@xiantang can you accept this PR? This is such an easy fix

Issue #516, opened 2024-01-22
".air.toml" fails to parse and air proceeds silently 


matheusgomes28:
When a user provides an invalid toml config that fails to be parsed, `air` will silently continue with default settings. Very confusing if you are debugging what's going on.

Example of wrong `.air.toml` config:

```toml
root = "."
testdata_dir = "testdata"
tmp_dir = "tmp"

[build]
  cmd = "make build-debug"
  args_bin = []
  bin = "./tmp/urchin-debug"
  full_bin = "dlv exec ./tmp/urchin-debug --listen=127.0.0.1:2345 --headless=true --api-version=2 --accept-multiclient --continue --log -- "
  delay = 1000
  exclude_dir = ["assets", "tmp", "vendor", "testdata"]
  exclude_file = []
  exclude_regex = ["_test.go"]
  exclude_unchanged = false
  follow_symlink = false
  include_dir = []
  include_ext = ["go", "tpl", "tmpl", "html"]
  include_file = []
  kill_delay = "0s"
  log = "build-errors.log"
  poll = false
  poll_interval = 0
  post_cmd = []
  full_bin = ""
  pre_cmd = []
  rerun = false
  rerun_delay = 500
  send_interrupt = false
  stop_on_error = false

[color]
  app = ""
  build = "yellow"
  main = "magenta"
  runner = "green"
  watcher = "cyan"

[log]
  main_only = false
  time = false

[misc]
  clean_on_exit = false

[screen]
  clear_on_rebuild = false
  keep_scroll = true
``` 

**Note** the multiple defined `full_bin`. For clarity, I was wondering why the following `Makefile` wasn't running properly:

```make
BINARY_NAME=urchin
URCHIN_BIN_PATH=./cmd/$(BINARY_NAME)

clean:
	rm -f ./bin/*

build-debug: clean
	echo "Hello There"
	CGO_ENABLED=0 go build -gcflags=all="-N -l" -o ./tmp/$(BINARY_NAME)-debug $(URCHIN_BIN_PATH)
```

Output from air with the above setup (simply typing `air` from dir with the `.air.toml` file):

```sh
$ air
 __    _   ___  
 / /\  | | | |_) 
/_/--\ |_| |_| \_ v1.49.0, built with Go go1.21.6
watching .

watching app
watching cmd
watching cmd/urchin
watching common
watching database
watching migrations
watching static
watching templates
watching templates/contact
watching templates/index
watching templates/post
watching templates/services
!exclude tmp
building...

[...]

PR #517, opened 2024-01-23
Adding Extra Logs for Config Reading Behaviour

matheusgomes28:
Improves the logs for #516 . After looking into the codebase, I understand that the following behaviour is intended:

1. read `.air.toml` if available.
2. failing above, read `.air.conf` if available.
3. failing above, use default configs.

The issue is that the user currently has no idea if they simply run `air` from their repo, with an invalid `.air.toml`. It is 100% clearer to log the behaviour so the user doesn't waste time (like I did) wondering why my `.air.toml` settings weren't picked up !

### DIFF
--- a/runner/config.go
+++ b/runner/config.go
@@ -176,15 +176,18 @@ func writeDefaultConfig() (string, error) {
 
 func defaultPathConfig() (*Config, error) {
 	// when path is blank, first find `.air.toml`, `.air.conf` in `air_wd` and current working directory, if not found, use defaults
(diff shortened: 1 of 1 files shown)


matheusgomes28:
Is air still under development? I can close PR if you guys aren't taking any more features

cosmtrek:
@matheusgomes28 Thanks for your PR. Air is indeed still being maintained, though it is currently not in a phase of active development. We will periodically address issues and pull requests in batches.

PR #518, opened 2024-01-26
Support creating parent directories as needed

mcieno:
I oftentimes find myself needing a multilevel folder structure in the `tmp` folder (e.g., `tmp/cmd1/...`, `tmp/cmd2/...`).

Unfortunately, `tmp_dir = tmp/cmd1` fails unless `tmp/` already exists.
Although this is very easy to work around by creating the parent directories beforehand, I believe that Air should be able to recursively create parents by default.

This change is equivalent to running `mkdir -p "$tmp_dir"` instead of `mkdir "$tmp_dir"`.


xiantang:
hello I think it already on the master now

mcieno:
Superseded by https://github.com/air-verse/air/pull/846.

PR #519, opened 2024-01-30
Add nix support

acaloiaro:
This PR allows `air` to be a Flake input for Nix users who use `air` in their Go projects. 

It also better supports `air` development by providing a development shell that contains all of `air`'s developer dependencies. 

I understand if this is not a priority for this project, and I'll continue maintaining my fork of air at `acaloiaro/air` if you choose not to directly support nix.

Cheers!


cosmtrek:
@acaloiaro Thanks! Looks great, while I need some time to learn Nix firstly :smile:

acaloiaro:
> @acaloiaro Thanks! Looks great, while I need some time to learn Nix firstly 😄

Sounds good @cosmtrek. I'm happy to clarify anything if you have questions.

One thing worth considering for this PR is that it serves two uses cases: users and developers.

For users it makes `air` available as a "flake input" for nix projects that use `air`. E.g. this allows Nix Flake users to delcare `air` as an input to their system configurations (make it globablly available), or as an input to projects, so that `air` is available within that project's "development shell".

For `air` devlopers who are also nix users, it allows them to simply run `nix develop` to get a development shell with all of `air`'s dependencies. 

I joined these two user and developer use cases into a single PR, but they could be easy considered in isolation. Selfishly, as a nix user, I want both. But it would be reasonable as a project maintainer to consider them separately.

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/air-verse/air
 
-go 1.23
```
xiantang, line 3: why do u change this?

acaloiaro: Because I forgot this PR was open. I use my fork regularly and update it as needed. 

I'm closing this PR for now, as there doesn't seem to be much demand for it. 

I'll reopen it if requested.

xiantang:
seems is not out of box
```
[neo@nixos:~/project/air]$ nix develop
error:
       while calling the 'derivationStrict' builtin

         at /builtin/derivation.nix:9:12: (source not available)

        while evaluating derivation 'devenv-shell'
         whose name attribute is located at /nix/store/y9m0i9f9nk27kgiwwm0nnc72bzr1p55j-source/src/modules/mkNakedShell.nix:63:12

       while evaluating attribute 'shellHook' of derivation 'devenv-shell'

         at /nix/store/y9m0i9f9nk27kgiwwm0nnc72bzr1p55j-source/src/modules/mkNakedShell.nix:78:5:

           77|     # requires that other trampoline.
           78|     shellHook = ''
             |     ^
           79|       ${lib.optionalString debug "set -x"}

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: devenv was not able to determine the current directory.
       Make sure Nix runs with the `--impure` flag.

       See https://devenv.sh/guides/using-with-flakes/
```

acaloiaro:
No demand for this feature.

PR #535, opened 2024-02-23
Redirect to browser after build

ValeryVerkhoturov:
`AppUrl` is used to open browser URL when the build is finished. `AppUrlOpenDelay` waits when server is ready to accept requests.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,6 +20,7 @@ require (
 	github.com/mattn/go-colorable v0.1.13 // indirect
 	github.com/mattn/go-isatty v0.0.17 // indirect
 	github.com/pelletier/go-toml/v2 v2.0.6 // indirect
+	github.com/pkg/browser v0.0.0-20240102092130-5ac0b6a4141c // indirect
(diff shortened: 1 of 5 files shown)


runner/engine.go
```diff
@@ -2,6 +2,7 @@ package runner
 
 import (
 	"fmt"
+	"github.com/pkg/browser"
```
RedCrazyGhost, line 5: You are advised to change the dependency location according to the following rules:
1. Official dependence.
2. External dependencies
3. Internal dependencies

go.mod
```diff
@@ -20,6 +20,7 @@ require (
 	github.com/mattn/go-colorable v0.1.13 // indirect
 	github.com/mattn/go-isatty v0.0.17 // indirect
 	github.com/pelletier/go-toml/v2 v2.0.6 // indirect
+	github.com/pkg/browser v0.0.0-20240102092130-5ac0b6a4141c // indirect
```
RedCrazyGhost, line 23: It seems to me that this project package, which has not been maintained for quite some time, may need to be considered for using this package

PR #551, opened 2024-03-20
Fix color output on Windows

### LINKED ISSUE #128: Colour of gin's log
How can I remove ?[97;42m] and change it to color?
I have already used gin.DefaultWriter = colorable.NewColorableStdout()
Goland can show color on gin's log, but no color when the system is launched by air.

launched by air
image

launched by goland
image

launched by .exe
image

8LWXpg:
Fix #128 

I simply compare `util_windows.go` with `util_unix.go`, and setting `c.Stdout`, `c.Stderr` seems to have resolve this issue.

## Screenshot

image

### DIFF
--- a/runner/util_windows.go
+++ b/runner/util_windows.go
@@ -2,6 +2,7 @@ package runner
 
 import (
 	"io"
+	"os"
 	"os/exec"
 	"strconv"
 	"strings"
@@ -29,6 +30,10 @@ func (e *Engine) startCmd(cmd string) (*exec.Cmd, io.ReadCloser, io.ReadCloser, 
 	if err != nil {
 		return nil, nil, nil, err
 	}
+
+	c.Stdout = os.Stdout
+	c.Stderr = os.Stderr
+
 	err = c.Start()
 	if err != nil {
 		return nil, nil, nil, err


8LWXpg:
A question for this code:
Returning `stdout` and `stderr` then `io.Copy` seems redundant, since setting `c.Stdout = os.Stdout` and `c.Stderr = os.Stderr` basically does the same thing. Is there reason for this?

cosmtrek:
@8LWXpg yes, the related code need to refactor 😄

runner/util_windows.go
```diff
@@ -29,6 +30,10 @@ func (e *Engine) startCmd(cmd string) (*exec.Cmd, io.ReadCloser, io.ReadCloser,
 	if err != nil {
 		return nil, nil, nil, err
 	}
+
+	c.Stdout = os.Stdout
+	c.Stderr = os.Stderr
```
makiuchi-d, line 35: These are overwrite StdoutPipe and StderrPipe.
https://cs.opensource.google/go/go/+/refs/tags/go1.23.1:src/os/exec/exec.go;l=1067

These pipe calls above should be removed.

PR #553, opened 2024-03-22
Fix go directive to avoid language version

haruyama480:
Specifying 1.22 (referred to as the language version) without the toolchain directive causes compilation to fail when developers uses older go like 1.21.
To workaround, I recommend to specify `go 1.22.0`.

https://github.com/golang/go/issues/62278#issuecomment-2015737150

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/cosmtrek/air
 
-go 1.22
+go 1.22.0
 
 require (
 	dario.cat/mergo v1.0.0


haruyama480:
https://github.com/cosmtrek/air/pull/529#discussion_r1493807604

Yes, the following error is local issue.

```
go: downloading go1.22 (darwin/arm64)
go: download go1.22 for darwin/arm64: toolchain not available
```

However, this merge works well without above error.

Issue #546, opened 2024-03-13
No such file or directory when project path has space


jarielbalberona:
I know that having spaces in folder names is weird but I have no problem with this setup with my previous projects and stacks.

/bin/sh: /Users/jariel/Documents/projects/1.: No such file or directory

I've fixed this by removing space in my folder name, from /1. dev/ to /1.dev/.

luisgarciaalanis:
you can also esape the space?

like so -> /1.\ dev

jarielbalberona:
I've no control over the path when I run the "air".

mateusfmello:
My directory structure has spaces because I use client company names to organize files, so some clients have names containing the space character

jonasricardo:
same problem. Any forecast for adjustment?

nanvenomous:
I'm having the same issue **except** my binary file is built as a https://github.com/spf13/cobra command.

my config:
```toml
  bin = "./main serve"
```
From docker container I get the same error message:
```
/bin/sh: line 1: /root/app/main serve: No such file or directory
```

My temporary fix is to cap my version of air (v1.61.7 doesn't have the bug):
```bash
go install github.com/air-verse/air@v1.61.7
```

nanvenomous:
looks like could be related to https://github.com/air-verse/air/pull/749 which was merged recently to add support for spaces in filepaths, but now cannot handle spaces in commands. @pkashin

williambao:
> I'm having the same issue **except** my binary file is built as a https:github.com/spf13/cobra command.
>
> my config:
>
>   bin = "./main serve"
> From docker container I get the same error message:
>
> ```
> /bin/sh: line 1: /root/app/main serve: No such file or directory
> ```
>
> My temporary fix is to cap my version of air (v1.61.7 doesn't have the bug):我的临时解决方法是限制我的空气版本（v1.61.7 没有这个错误）：
>
> go install github.com/air-verse/air@v1.61.7

Yes, I have the same problem. Version 1.62.0 cannot start my app (also using cobra). After downgrading to version 1.61.7, there is no problem.

nanvenomous:
I figured out how to resolve this on later versions of air!
You must use `args_bin` configuration.
```toml
[build]
bin = './main'
args_bin = ['serve']
``` 

which is equivalent to the following shell command:
```bash
./main serve
```

xiantang:
see this issue https://github.com/air-verse/air/issues/824

xiantang:
[...]

PR #199, opened 2021-10-18
Replace space in folder name with backslash

TheYkk:
Signed-off-by: Kaan Karakaya <yusufkaan142@gmail.com>

PR #570, opened 2024-04-15
fix: escape spaces in bin

wobww:
Air was unable to run my binary because there were spaces in the path, the same issue is described here -> #546.

This change fixes it by escaping spaces in the path at config setup time.

### DIFF
--- a/runner/config.go
+++ b/runner/config.go
@@ -9,6 +9,7 @@ import (
 	"reflect"
 	"regexp"
 	"runtime"
+	"strings"
 	"time"
 
 	"dario.cat/mergo"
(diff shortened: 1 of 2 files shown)


ainsleyclark:
Hi, will this be merged soon? Noticed this problem when having a space in the folder name.

osteensco:
Bumping this as I just started trying to use air and had the same issue that this PR solves.

makiuchi-d:
Seems to be a duplicate of #199.
And, it doesn't work on windows.

osteensco:
> Seems to be a duplicate of #199. And, it doesn't work on windows.

I submitted #646, which should work on windows as well.

xiantang:
https://github.com/air-verse/air/pull/646 is merged, close this pr. thanks

PR #585, opened 2024-05-14
fix: fix proxy error handling and race conditions

ndajr:
I improved the error handling by avoiding log.Fatal as much as possible which doesn't clean up the resources properly (leaving the process and ports open). I couldn't reproduce [this issue](https://github.com/cosmtrek/air/issues/584) with the steps mentioned but after running `go test -v -count=1 -race -run TestProxyStream ./runner` a few times I got some race condition warnings. You can run those tests to validate the new behaviour (with atomic counters and RWMutex) working as expected with no warnings. I also added a CORS wildcard allow asked in another issue.

### DIFF
--- a/runner/proxy_stream.go
+++ b/runner/proxy_stream.go
@@ -6,7 +6,7 @@ import (
 )
 
 type ProxyStream struct {
-	mu          sync.RWMutex
+	mu          sync.Mutex
 	subscribers map[int32]*Subscriber
 	count       atomic.Int32
 }
--- a/runner/proxy_stream.go
+++ b/runner/proxy_stream.go
@@ -2,45 +2,49 @@ package runner
 
 import (
 	"sync"
+	"sync/atomic"
 )
 
 type ProxyStream struct {
-	sync.Mutex
-	subscribers map[int]*Subscriber
-	count       int
+	mu          sync.RWMutex
+	subscribers map[int32]*Subscriber
+	count       atomic.Int32
 }
 
 type Subscriber struct {
-	id       int
+	id       int32
 	reloadCh chan struct{}
 }
 
 func NewProxyStream() *ProxyStream {
-	return &ProxyStream{subscribers: make(map[int]*Subscriber)}
+	return &ProxyStream{subscribers: make(map[int32]*Subscriber)}
 }
 
 func (stream *ProxyStream) Stop() {
 	for id := range stream.subscribers {
 		stream.RemoveSubscriber(id)
 	}
-	stream.count = 0
+	stream.count = atomic.Int32{}
 }
 
 func (stream *ProxyStream) AddSubscriber() *Subscriber {
-	stream.Lock()
-	defer stream.Unlock()
-	stream.count++
+	stream.mu.Lock()
+	defer stream.mu.Unlock()
+	stream.count.Add(1)
 
-	sub := &Subscriber{id: stream.count, reloadCh: make(chan struct{})}
-	stream.subscribers[stream.count] = sub
+	sub := &Subscriber{id: stream.count.Load(), reloadCh: make(chan struct{})}
+	stream.subscribers[stream.count.Load()] = sub
 	return sub
 }
 
-func (stream *ProxyStream) RemoveSubscriber(id int) {
-	stream.Lock()
-	defer stream.Unlock()
-	close(stream.subscribers[id].reloadCh)
-	delete(stream.subscribers, id)
+func (stream *ProxyStream) RemoveSubscriber(id int32) {
+	stream.mu.Lock()
+	defer stream.mu.Unlock()
+
+	if _, ok := stream.subscribers[id]; ok {
+		close(stream.subscribers[id].reloadCh)
+		delete(stream.subscribers, id)
+	}
 }
 
 func (stream *ProxyStream) Reload() {
--- a/runner/proxy_stream_test.go
+++ b/runner/proxy_stream_test.go
@@ -2,10 +2,11 @@ package runner
 
 import (
 	"sync"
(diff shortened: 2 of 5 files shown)


xiantang:
hey, coverage is required. currently I wanna use `"github.com/bytedance/mockey"` to mock the filesystem event. because the old style ut is unstable.

xiantang (approved):

runner/proxy_stream.go
```diff
@@ -2,45 +2,49 @@ package runner
...
-	close(stream.subscribers[id].reloadCh)
-	delete(stream.subscribers, id)
+func (stream *ProxyStream) RemoveSubscriber(id int32) {
+	stream.mu.Lock()
+	defer stream.mu.Unlock()
+
+	if _, ok := stream.subscribers[id]; ok {
+		close(stream.subscribers[id].reloadCh)
+		delete(stream.subscribers, id)
+	}
```
xiantang, line 47: need print log when we didn't find subscriber

xiantang:
and i'm not sure, Is this panic caused by race conditions?

runner/proxy_stream.go
ndajr: We only have `mainLog` on the Engine struct, we would need to pass it down to Proxy and ProxyStream. I think it's not necessary, I added this just as an extra precaution to avoid nil pointer issues. Missing the subscriber in the map means it's already cleaned up, and during air shutdown we clean up all subscribers, so it's all taken care of. If you still think we should pass down the logger to log here I can do it

ndajr:
This is a normal flow
1. the user runs air with proxy enabled and access the proxy via browser
2. the script is injected on the page
3. the browser connects to `/internal/reload` endpoint (server-sent events)
4. A new subscriber is created on proxy_stream
5. The connection with /internal/reload stays alive until the user closes the tab, causing the connection to be closed (context Done signal is sent) and we remove the subscriber in the subscribers map

If the user opened two tabs accessing the proxy, we would have 2 subscribers, and if suddenly the browser is closed those two subscriptions will be removed from the map, but a map is not thread safe, that's why I am moving from sync.Mutex (which prevents race condition with reads) to `sync.RWMutex` (prevents race condition with reads and writes), and also using an atomic counter which is recommended when dealing with multiple go routines (https://gobyexample.com/atomic-counters). I hope this gives an explanation on why I think the current code can cause panics when a race condition happens and I could validate it running the tests against the master branch and this branch (`go test -v -count=1 -race -run TestProxyStream ./runner`).

runner/proxy_stream.go
xiantang: ic, no need to add

runner/proxy_stream.go
```diff
@@ -2,45 +2,49 @@ package runner
...
 import (
 	"sync"
+	"sync/atomic"
 )
 
 type ProxyStream struct {
-	sync.Mutex
-	subscribers map[int]*Subscriber
-	count       int
+	mu          sync.RWMutex
```
xiantang, line 9: I think you are using wrong, as I know RWMutex is **Allow Multiple Readers**. but it will be block when u try to write.

call RLock, when u  are reading a resource.

ndajr: What do you mean by reading a resource? We don't need the mutex when reading from the reload channel, see here: https://github.com/cosmtrek/air/blob/master/runner/proxy.go#L169. We need the lock when we modify the subscriptions map. So we can create as many subscribers as we want and SSE endpoint will provide updates for all of them in parallel, they will never block each other if that's your concern

ndajr:
Regarding the coverage, I will see if I can improve the tests for proxy. But for `engine.go` it decreased -0.17% just by checking the proxy error and failed the job. Is there a way we can increase the coverage threshhold? Otherwise I will have to add tests for functions that are not proxy related

runner/proxy_stream.go
xiantang: > We don't need the mutex when reading from the reload channel
so we don't need to use RWMutex, just use mutex is fine. I did see any benefit by using this RWMutex.

the reason is AddSubscriber and RemoveSubscriber both modify the `subscribers` map. 
in my opinion,  RWMutex is going to improve the reading performance.  
can see this answer https://stackoverflow.com/a/19168242

xiantang:
> Regarding the coverage, I will see if I can improve the tests for proxy. But for `engine.go` it decreased -0.17% just by checking the proxy error and failed the job. Is there a way we can decrease the minimum coverage threshhold? Otherwise I will have to add tests for functions that are not proxy related

UT is fine. because u already covered enough

runner/proxy_stream_test.go
```diff
@@ -28,39 +29,43 @@ func TestProxyStream(t *testing.T) {
...
 	}()
 
-	reloadCount := 0
+	var reloadCount atomic.Int32
 	for _, sub := range stream.subscribers {
 		wg.Add(1)
 		go func(sub *Subscriber) {
 			defer wg.Done()
 			<-sub.reloadCh
-			reloadCount++
```
xiantang, line 44: this is the reason why u got race condition

runner/proxy_stream.go
```diff
@@ -2,45 +2,49 @@ package runner
...
 	reloadCh chan struct{}
 }
 
 func NewProxyStream() *ProxyStream {
-	return &ProxyStream{subscribers: make(map[int]*Subscriber)}
+	return &ProxyStream{subscribers: make(map[int32]*Subscriber)}
 }
 
 func (stream *ProxyStream) Stop() {
 	for id := range stream.subscribers {
```
xiantang, line 24: we need to lock here.

PR #586, opened 2024-05-14
Fix: Proxy return redirect response

reginbald:
## What has been changed
Configures proxy to not follow redirects and return to the client the actual response from the server.

## Why?
Current proxy solution "follows" redirect responses from the server and makes an additional request rather then returning the redirect response to the client.

## Example Scenario
Client issues a delete request and the server returns a See Other redirect response. Current proxy solution does not return the response but follows the redirect response, makes another request and then returns that response to the client.

### Before the change:
image
image

### After the change:
image
image

### DIFF
--- a/runner/proxy.go
+++ b/runner/proxy.go
@@ -33,7 +33,11 @@ func NewProxy(cfg *cfgProxy) *Proxy {
 		server: &http.Server{
 			Addr: fmt.Sprintf(":%d", cfg.ProxyPort),
 		},
-		client: &http.Client{},
+		client: &http.Client{
+			CheckRedirect: func(req *http.Request, via []*http.Request) error {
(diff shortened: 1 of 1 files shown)


reginbald:
Hey @ndajr do you have time to review this PR? Any thoughts?

ndajr:
Hi @reginbald, sorry for the delay. Looks good to me! I will try to reach @xiantang and see if we can release these three PRs together in the next minor release:
- https://github.com/cosmtrek/air/pull/585
- https://github.com/cosmtrek/air/pull/586
- https://github.com/cosmtrek/air/pull/587

xiantang (approved):

xiantang:
Tks. well done

PR #592, opened 2024-05-28
fix: add golangci-lint setup action and fix lint issue

ndajr:
The `make check` linting step failed: https://github.com/cosmtrek/air/actions/runs/9240465156/job/25420774320. The build step running in each PR is broken with `golangci-lint: command not found` issue: https://github.com/cosmtrek/air/actions/runs/9149739894/job/25364966455#step:4:13

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,11 +20,14 @@ jobs:
         uses: actions/setup-go@v4
         with:
           go-version: ^1.22
-
-      - name: Build
-        run: make build
+      - name: Install golangci-lint
+        uses: golangci/golangci-lint-action@v6
+        with:
+          version: v1.59
       - name: Install dependency
         run: if [ $(uname) == "Darwin" ]; then brew install gnu-sed ;fi
+      - name: Build
+        run: make build
       - name: Run unit tests
         run: go install github.com/go-delve/delve/cmd/dlv@latest && go test ./... -v -covermode=count -coverprofile=coverage.txt
       - name: Upload Coverage report to CodeCov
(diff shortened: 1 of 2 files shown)


xiantang (approved):

.github/workflows/build.yml
```diff
@@ -20,11 +20,14 @@ jobs:
         uses: actions/setup-go@v4
         with:
           go-version: ^1.22
-
-      - name: Build
-        run: make build
+      - name: Install golangci-lint
+        uses: golangci/golangci-lint-action@v6
+        with:
+          version: v1.59
```
ccoVeille, line 26: Tip: You could have used "latest" so you don't have to maintain it

.github/workflows/build.yml
```diff
@@ -20,11 +20,14 @@ jobs:
         uses: actions/setup-go@v4
         with:
           go-version: ^1.22
```
ccoVeille, line 22: Not related to your change but related to my other comment

https://github.com/cosmtrek/air/pull/592#discussion_r1617032972

Here you can use "stable" it will always use the latest version and you don't have to maintain it

ndajr: We could but it has downsides as well. It may be "easier" to maintain but may break all the users of the project, which has a bigger impact than just change the version in every few months. See https://vsupalov.com/docker-latest-tag/

ccoVeille: > We could but it has downsides as well. It may be "easier" to maintain but may break all the users of the project, which has a bigger impact than just change the version in every few months. See https://vsupalov.com/docker-latest-tag/

Very good read, I get it.

.github/workflows/build.yml
halkeye: fyi, Same reply from the other comment, but i'll throw in tools like renovate exist so you can automate updating things, while keeping consistent build experience.

ccoVeille: I'm sure dependabot cannot touch this argument of an action. It can bump the action version. But not its argument.

But maybe renovate can do it, so it would be fine

PR #594, opened 2024-05-28
ci: add GitHub dependabot

ccoVeille:
It will be incharge of opening PR if a Go dependency or a GHA should be updated

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,21 @@
+version: 2
+updates:
+  # Maintain golang dependencies defined in go.mod
+  # These would open PR, these PR would be tested with the CI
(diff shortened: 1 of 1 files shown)


ccoVeille:
I noticed setup-go was in version 4 in GHA

https://github.com/cosmtrek/air/blob/e3b6eafe8afd4ac8640421bad573ea5e655592cd/.github/workflows/build.yml#L20

while the last version is version 5 see https://github.com/marketplace/actions/setup-go-environment

I looked at existing issues or PR and found no mention of dependabot/renovate, so I made a PR to add dependabot to repository,

ccoVeille:
Any raison for this simple PR to be left aside?

PR #598, opened 2024-06-02
add support for environment files

mpldr:
At times, it is necessary or at least easier to set environment variables using a file.

Add a function that augments the environment of the spawned process with the provided variables.

Tested to work on Linux. Not sure on how to write a test for it without either passing in the content as a reader or basically testing stdlib IO functions.

### DIFF
--- a/runner/environment.go
+++ b/runner/environment.go
@@ -1,0 +1,30 @@
+package runner
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"os/exec"
+)
+
+func (e *Engine) modifyEnvironment(c *exec.Cmd) error {
+	if e.config.Build.EnvironmentFile == "" {
+		return nil
+	}
+
+	env := os.Environ()
+
+	envFile, err := os.Open(e.config.Build.EnvironmentFile)
+	if err != nil {
+		return fmt.Errorf("failed to open environment file: %w")
+	}
+	defer envFile.Close()
+
+	scan := bufio.NewScanner(envFile)
+	for scan.Scan() {
+		env = append(env, scan.Text())
+	}
+
+	c.Env = env
+	return nil
+}
--- a/runner/config.go
+++ b/runner/config.go
@@ -58,6 +58,7 @@ type cfgBuild struct {
 	KillDelay        time.Duration `toml:"kill_delay"`
 	Rerun            bool          `toml:"rerun"`
 	RerunDelay       int           `toml:"rerun_delay"`
+	EnvironmentFile  string        `toml:"environment_file"`
 	regexCompiled    []*regexp.Regexp
 }
 
--- a/runner/util_linux.go
+++ b/runner/util_linux.go
@@ -30,6 +30,10 @@ func (e *Engine) killCmd(cmd *exec.Cmd) (pid int, err error) {
 
 func (e *Engine) startCmd(cmd string) (*exec.Cmd, io.ReadCloser, io.ReadCloser, error) {
 	c := exec.Command("/bin/sh", "-c", cmd)
+	err := e.modifyEnvironment(c)
+	if err != nil {
+		return nil, nil, nil, err
+	}
 	f, err := pty.Start(c)
 	return c, f, f, err
 }
--- a/runner/util_unix.go
+++ b/runner/util_unix.go
@@ -33,6 +33,10 @@ func (e *Engine) startCmd(cmd string) (*exec.Cmd, io.ReadCloser, io.ReadCloser, 
 	c.SysProcAttr = &syscall.SysProcAttr{
 		Setpgid: true,
 	}
+	err := e.modifyEnvironment(c)
+	if err != nil {
+		return nil, nil, nil, err
+	}
 
 	stderr, err := c.StderrPipe()
 	if err != nil {
--- a/runner/util_windows.go
+++ b/runner/util_windows.go
@@ -25,6 +25,10 @@ func (e *Engine) startCmd(cmd string) (*exec.Cmd, io.ReadCloser, io.ReadCloser, 
 	if err != nil {
 		return nil, nil, nil, err
 	}
+	err = e.modifyEnvironment(c)
+	if err != nil {
+		return nil, nil, nil, err
+	}
 	stdout, err := c.StdoutPipe()
 	if err != nil {
 		return nil, nil, nil, err


runner/environment.go
```diff
@@ -0,0 +1,27 @@
+package runner
+
+import (
+	"bufio"
+	"os"
+	"os/exec"
+)
+
+func (e *Engine) modifyEnvironment(c *exec.Cmd) {
```
ccoVeille, line 9: Side remark, but is the "bare return" you apply in this method a common thing in this repository

revive discourages using it, and I agree with them

https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#bare-return

runner/environment.go
```diff
@@ -0,0 +1,27 @@
...
+	env := os.Environ()
+
+	envFile, err := os.Open(e.config.Build.EnvironmentFile)
+	if err != nil {
+		return
+	}
+
+	scan := bufio.NewScanner(envFile)
+	for scan.Scan() {
+		env = append(env, scan.Text())
```
ccoVeille, line 23: Won't you import empty line or line with comment?

Is there any library you could have used?

runner/environment.go
mpldr: It might be a better idea to return an error, you're right.

runner/environment.go
mpldr: Empty lines, like all lines without `=` are ignored, lines like `# this is a comment = true` add a very strange variable to the environment. Neither should pose much of a problem.

runner/environment.go
ccoVeille: By bare return, I meant using `return c`everywhere and not `return`

runner/environment.go
ccoVeille: Yes, they should be ignored. My point is what is the code that is not in this PR, that skip them

runner/environment.go
ccoVeille: But your new implementation is better, I like it 👍

mpldr:
On Sun 02 Jun 2024 05:11:53, ccoVeille wrote:
> +func (e *Engine) modifyEnvironment(c *exec.Cmd) {
>
> By bare return, I meant using `return c`everywhere and not `return`

That is completely missing the point of a bare return, I think. If the
function signature does not specify any returns, of course a bare return
is the way to… well… return. Bare returns, at least as I understand it
are referring to:

```go
// bare return, i. e. bad
func (e *Engine) modifyEnvironment(c *exec.Cmd) (err error) {
    err = errors.New("I feel bare")
    return
}

// non-bare return, i. e. better
func (e *Engine) modifyEnvironment(c *exec.Cmd) error {
    return errors.New("I feel bare")
}
```

To be fair, not the best example, but it gets the point across. Since it
wasn't using a bare return this doesn't apply. I am still happy to
report that the added return is not bare :)

-- 
Moritz Poldrack
https://moritz.sh

> Not intended to diagnose, treat, cure or prevent any disease.

mpldr:
OS handling of environment variables and Go stdlib.

-- 
Moritz Poldrack
https://moritz.sh

> No alcohol, dogs or horses.

xiantang:
why do u just pass env var like this:

`A=B air`

ccoVeille:
> why do u just pass env var like this:
> 
> `A=B air`

@xiantang I was also wondering for me the environment variables are what they are supposed to be something global that exists prior the binary is launched, they are added by the shell.

I was surprised by what @mpldr added, but I was like, never mind 🤷, but as you raise the point 🤨 too. I feel like my first thought was right.

I wouldn't expect this to be added to `air`

It's usually the role of:
- a runner tool that will import the variables, then launch air
- a shell script
- a global environment variable added to the shell
- any virtualenv tool
- a command-line scoped environment variable as you spotted @xiantang

mpldr:
> why do u just pass env var like this:
> 
> `A=B air`

mainly two reasons:
- No way to change the environment between restart, which at times can be necessary (for example when changing parsing logic or moving on to a different task)
- This gets ridiculously long for some docker services, where you have to set a dozen or more environment variables.

> they are supposed to be something outside the binary that exists prior the binary is launched

Since air acts as the launcher (it calls `sh`), this seems like an appropriate place to do it, to me.

ccoVeille:
What about using the air config file to set the env variables, so not the location of an env file?

mpldr:
That was an approach I was considering, but ultimately abandoned it because .env files or some variation of them are usually already present. And having two places to maintain environment variables might be a tough sell :)

ccoVeille:
Thanks for your reply.

It won't change the points I already raised, it's surprising to see such implementation.


https://github.com/air-verse/air/pull/598#issuecomment-2144833248

For me, it's over the role of an app to do that

I look at what supervisord was doing and it was providing such a feature and the reply and arguments were very similar

https://github.com/Supervisor/supervisor/issues/771


But anyway, why not… but I would discourage you doing it.

ccoVeille:
I'm still awaiting a reply from this

https://github.com/air-verse/air/pull/598#discussion_r1623397536

Because for me, with current code, it would import line as they are so this:

# whatever
FOO=42 # bar

Would case problems I think

runner/environment.go
mpldr: Sorry, GitHub's lovely handling of email replies put this in the main thread.

https://github.com/air-verse/air/pull/598#issuecomment-2143825842

Otherwise it would be a simple if-statement. Still no need to import anything :) Let me know if I should add that clause.

ccoVeille: Yes,i was expecting an if in the scan look

Maybe something like

```go
line = strings.Split(line, "#)
line = strings.TrimSpace(line)

if line == "" {
     continue
}
```

ccoVeille: > Sorry, GitHub's lovely handling of email replies put this in the main thread.
> 
>


> OS handling of environment variables and Go stdlib.

 That's my point. You are not using Go stdlib when handling the variables the way you do in this PR

> Otherwise it would be a simple if-statement. Still no need to import anything :) Let me know if I should add that clause.

Yes,i was expecting an if in the scan loop

Maybe something like

```go
line = strings.Split(line, "#)
line = strings.TrimSpace(line)

if line == "" {
     continue
}
```

ccoVeille: I mean you are using os.Environ. so yes, it skip the comments, empty lines. 

But your code doesn't when you add the lines of the env file

ccoVeille: I'm worried that you would have to code everything needed to do things like this

```env
BAR=42
FOO=BAR
```

Your code will set too to BAR not 42

I'm also worried that you would have to comply with uncommon but yet still used emv syntax like this

```env

FOO=${BAR:-42} # If BAR not set or null, use 42.
```

I might be totally wrong, but from my perspective this should be at least covered by tests

ccoVeille: @mpldr @xiantang

ccoVeille: I found backfound the project I was looking for

https://github.com/joho/godotenv

I think you should reluy on a lib like this to load and parse the env file, then you can append to existing env

ccoVeille:
I'm still not convinced, but it's not a problem 😅

But if you want to keep the feature please look at https://github.com/air-verse/air/pull/598#discussion_r1625432195

runner/environment.go
mpldr: > I'm worried that you would have to code everything needed to do things like this
> 
> ```dotenv
> BAR=42
> FOO=BAR
> ```
> 
> Your code will set too to BAR not 42

That's the expected behaviour, but you're right, it doesn't account for any kind of shell interpretation. I'll have a look at the suggested library later today. :)

PR #611, opened 2024-06-06
fix: fix proxy unmodified response when body tag is missing

ndajr:
Based on https://github.com/air-verse/air/pull/604. I found other issues as well, we are missing some returns in case of http internal error responses


ndajr:
This coverage thing is a bit annoying, by adding returns it thinks the coverage reduced but it's in practise unchanged, we are just fixing the bug. Any ideas on how to fix this or easy ways to improve coverage?

PR #619, opened 2024-06-23
 Implement graceful shutdown on Mac 

### LINKED ISSUE #534: Implement graceful shutdown on Mac
I usually unconsciously do Ctrl-C and I found this what I consider a bug or maybe it could just because of my code. I'm not really sure. So when the server is up and running and everything goes perfect right? and when I exit with Ctrl-C it does not finish the exiting process causing port already in use on next start.

```
2024/02/23 10:50:34 ⇨ MongoDB Connected!

   ____    __
  / __/___/ /  ___
 / _// __/ _ \/ _ \
/___/\__/_//_/\___/ v4.11.4
High performance, minimalist Go web framework
https://echo.labstack.com
____________________________________O/_______
                                    O\
⇨ http server started on [::]:1323
^Ccleaning...
see you again~
❯ air

  __    _   ___
 / /\  | | | |_)
/_/--\ |_| |_| \_ v1.50.0, built with Go go1.22.0

watching .
watching cmd
watching cmd/aion
watching internal
watching internal/cmd
watching internal/db
watching internal/handlers
watching internal/middlewares
watching internal/models
watching internal/routes
watching internal/utils
!exclude tmp
building...
running...
2024/02/23 10:50:38 ⇨ MongoDB Connected!

   ____    __
  / __/___/ /  ___
 / _// __/ _ \/ _ \
/___/\__/_//_/\___/ v4.11.4
High performance, minimalist Go web framework
https://echo.labstack.com
____________________________________O/_______
                                    O\
{"time":"2024-02-23T10:50:38.557171+08:00","level":"FATAL","prefix":"echo","file":"api.go","line":"38","message":"listen tcp :1323: bind: address already in use"}
Process Exit with Code: 1
```

My code on graceful shutdown
```
	// Wait for interrupt signal to gracefully shutdown the server with a timeout of 10 seconds.
	// Use a buffered channel to avoid missing signals as recommended for signal.Notify
	quit := make(chan os.Signal, 1)
	signal.Notify(quit, os.Interrupt)
	<-quit
	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
	defer cancel()

	// Shutdown server
	if err := e.Shutdown(ctx); err != nil {
		e.Logger.Fatal(err)
	}

	// Disconnect from MongoDB
	if err := db.Client.Disconnect(context.Background()); err != nil {
		log.Fatal("⇨ Error disconnecting to MongoDB: ", err)
		log.Fatal(err)
	}
```

xiantang:
close https://github.com/air-verse/air/issues/534

### DIFF
--- a/runner/engine.go
+++ b/runner/engine.go
@@ -30,8 +30,8 @@ type Engine struct {
 	buildRunStopCh chan bool
 	binStopCh      chan bool
 	exitCh         chan bool
-	procKilledCh   chan bool
 
+	cleanupWg     sync.WaitGroup
 	mu            sync.RWMutex
 	watchers      uint
 	fileChecksums *checksumMap
@@ -59,7 +59,6 @@ func NewEngineWithConfig(cfg *Config, debugMode bool) (*Engine, error) {
 		buildRunStopCh: make(chan bool, 1),
 		binStopCh:      make(chan bool),
 		exitCh:         make(chan bool),
-		procKilledCh:   make(chan bool),
 		fileChecksums:  &checksumMap{m: make(map[string]string)},
 		watchers:       0,
 	}
@@ -480,6 +479,7 @@ func (e *Engine) runPostCmd() error {
 }
 
 func (e *Engine) runBin() error {
+	e.cleanupWg.Add(1)
 	killFunc := func(cmd *exec.Cmd, stdout io.ReadCloser, stderr io.ReadCloser, killCh chan struct{}, processExit chan struct{}, wg *sync.WaitGroup) {
 		defer wg.Done()
 		select {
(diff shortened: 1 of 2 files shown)


runner/engine.go
```diff
@@ -536,6 +537,7 @@ func (e *Engine) runBin() error {
 			case <-killCh:
 				return
 			default:
+				e.procKillWg.Add(1)
```
nakkamarra, line 540: What I don't understand is why the WaitGroup passed to `killFunc` on line 553 doesn't seem to be working properly, I'd expect that to have waited for the process to be killed.

xiantang: because killFunc is running in a goroutine

nakkamarra: No that's expected of course, but the WaitGroup is expects as an argument is defined on 522 and is only ever `Wait`ed on line 528, so perhaps waiting on that wasn't working as expected or the `Wait` wasn't being called every time. To me, at least, figuring that out seems like a better solution than adding yet another WaitGroup

xiantang: when i remove the wg.Done(), the we still can exit. i think it's pointer issue. because we put the pointer into funtion

nakkamarra (approved):

PR #629, opened 2024-07-24
Set the Via header in proxy requests to origin and in client responses

dedalusj:
Set the Via HTTP header in requests to origin app and in the client responses.

Applications that uses the oauth workflow require the server to redirect the user back to the callback address. However when running the application behind the air proxy there is no information about which host is the proxy running out. This causes the application to redirect the user to a callback path on the host the application is running at causing a breakdown of the oath flow.

Adding the Via header to the proxy requests allows the applications to detect the proxy host and craft an appropriate callback URL for the oauth flow.


cosmtrek:
@dedalusj Hi, could you fix the conflicts?

dedalusj:
@cosmtrek Done

PR #641, opened 2024-09-07
feat: silent flag in .air.toml, suppressing all air prints

jesses-code-adventures:
useful if the app being watched has structured outputs that are being piped into subsequent program/s.

### DIFF
--- a/air_example.toml
+++ b/air_example.toml
@@ -58,6 +58,8 @@ rerun_delay = 500
 time = false
 # Only show main log (silences watcher, build, runner)
 main_only = false
+# silence all logs produced by air 
+silent = false
 
 [color]
 # Customize each part's color. If no color found, use the raw app log.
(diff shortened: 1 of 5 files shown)


silverwind:
Will errors still show with this flag enabled? If so, it should be fine imho.

jesses-code-adventures:
this library generally uses log.Fatalf() for errors, whereas these changes only affect info and debug logs in runner/util.go and runner/logger.go which produce the lion's share of the logs. shouldn't be anything to worry about as far as errors being suppressed.

cosmtrek:
@jesses-code-adventures Hi, thanks for adding the silent flag, could you rebase the master and rerun the checks?

jesses-code-adventures:
@cosmtrek i have rebased the master but i don't believe i can trigger the checks on my end. please let me know if I'm wrong on this though, happy to add an explanation to the readme if contributors can run the checks.

Issue #472, opened 2023-10-11
Issue when running when inside a folder containing whitespace


AloizioMacedo:
It seems that having folders like

"/Documents/Go Projects/my_project"

presents problems, and air trims the example above to

"/Documents/Go"

and shows permission denied issues.

After I changed the name of the offending folder in the example above to "Go_Projects", everything worked nicely.

xiantang:
thank your report, let me reproduce it

xiantang:
may i know what OS do u use?

xiantang:
![image](https://github.com/cosmtrek/air/assets/34479567/46775611-a8a0-4777-953b-5da7d060b8eb)

xiantang:
https://github.com/golang/go/issues/17149 related

DevDorrejo:
have this issue with this name "Let's Go".

AloizioMacedo:
> may i know what OS do u use?

Ubuntu 20.04 inside WSL, Windows 11.

makiuchi-d:
I was fixed, but it was reverted by @xiantang.
https://github.com/cosmtrek/air/pull/179

jrmylua:
```
running...
CMD will not recognize non .exe file for execution, path: C:\Users\Jeremy Lua\Documents\code\jvbe\tmp\jvbe app
'C:\Users\Jeremy' is not recognized as an internal or external command,
operable program or batch file.
Process Exit with Code: 1
```
I have a similar issue with my project path. Running Windows 10 Education Version 22H2

ctholho:
I just wanted to mention that this is not a windows issue. I have the same problem running this on macos.

<img width="859" alt="Screenshot 2024-05-27 at 21 57 49" src="https://github.com/cosmtrek/air/assets/28910105/a6557f06-8b50-4281-b20c-b7ba20abe141">

xiantang:
should be fixed in latest

PR #646, opened 2024-09-19
FIX: Support working directories containing whitespace

osteensco:
addresses #546 #472, probably more

I believe this solution is OS agnostic but happy to adjust if it doesn't work somewhere.

### DIFF
--- a/runner/config.go
+++ b/runner/config.go
@@ -79,6 +79,7 @@ func (c *cfgBuild) RegexCompiled() ([]*regexp.Regexp, error) {
 type cfgLog struct {
 	AddTime  bool `toml:"time"`
 	MainOnly bool `toml:"main_only"`
+	Silent   bool `toml:"silent"`
 }
 
 type cfgColor struct {
@@ -187,7 +188,7 @@ func defaultPathConfig() (*Config, error) {
 	for _, name := range []string{dftTOML, dftConf} {
 		cfg, err := readConfByName(name)
 		if err == nil {
-			if name == dftConf {
+			if name == dftConf && !cfg.Log.Silent {
 				fmt.Println("`.air.conf` will be deprecated soon, recommend using `.air.toml`.")
 			}
 			return cfg, nil
@@ -238,6 +239,7 @@ func defaultConfig() Config {
 	log := cfgLog{
 		AddTime:  false,
 		MainOnly: false,
+		Silent:   false,
 	}
 	color := cfgColor{
 		Main:    "magenta",
--- a/runner/config.go
+++ b/runner/config.go
@@ -9,6 +9,7 @@ import (
 	"reflect"
 	"regexp"
 	"runtime"
+	"strings"
 	"time"
 
 	"dario.cat/mergo"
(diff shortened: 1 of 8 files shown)


osteensco:
sorry for the messy merges... I was having issues with some of the tests that seemed unrelated to my changes, and the build process on github kept failing so I tried to catchup since I saw there was a linter problem fixed. I ended up just testing my PR manually on a local project and it works on my machine 🤷‍♂️ . Is there a known issue with some of the tests in their current state? I couldn't find any open issues about it.

omarattia3143:
this is causing issue #659, I am not sure why does space in path an cause issues, test paths with spaces using windows 11 it works fine, so I am not sure how to test this but could you use something like this?
`if strings.Contains(c.Build.Bin, " ") && !strings.HasPrefix(c.Build.Bin, "\"") && !strings.HasSuffix(c.Build.Bin, "\"") {
		c.Build.Bin = "\"" + c.Build.Bin + "\""
	}`

cosmtrek:
@omarattia3143 Sorry for the bug, I don't have windows for testing, PR is welcome~

omarattia3143:
@osteensco can you confirm that this PR still works with this code? I have only windows and I can not test your changes

> this is causing issue #659, I am not sure why does space in path an cause issues, test paths with spaces using windows 11 it works fine, so I am not sure how to test this but could you use something like this? `if strings.Contains(c.Build.Bin, " ") && !strings.HasPrefix(c.Build.Bin, "\"") && !strings.HasSuffix(c.Build.Bin, "\"") { c.Build.Bin = "\"" + c.Build.Bin + "\"" }`

runner/config.go
```diff
@@ -327,6 +329,9 @@ func (c *Config) preprocess() error {
 	// CMD will not recognize relative path like ./tmp/server
 	c.Build.Bin, err = filepath.Abs(c.Build.Bin)
 
+	// Account for spaces in filepath
+	c.Build.Bin = fmt.Sprintf("%q", c.Build.Bin)
```
JasonGoemaat, line 333: I think this is causing an error on Windows, the exe can't be found and the path includes escape characters like this:

```
'\"c:\\go\\hello\\tmp\\hello.exe\"' is not recognized as an internal or external command,
operable program or batch file.
```

When if it was actually missing the error would be: this when the exe is missing and I run it from a command prompt
```
'c:\go\hello\tmp\hello.exe' is not recognized as an internal or external command,
operable program or batch file.
```

moboqe: yep, that is not working properly

moboqe:
> this is causing issue #659, I am not sure why does space in path an cause issues, test paths with spaces using windows 11 it works fine, so I am not sure how to test this but could you use something like this? `if strings.Contains(c.Build.Bin, " ") && !strings.HasPrefix(c.Build.Bin, "\"") && !strings.HasSuffix(c.Build.Bin, "\"") { c.Build.Bin = "\"" + c.Build.Bin + "\"" }`

That seems not working either

omarattia3143:
@moboqe this is not merged, It works if you dont have space in your path, I made it so it can bypass the issue created by this PR.
I am not sure if this PR works for other platforms.
@cosmtrek I think the best is to revert this PR since the author is not responding and it is affecting all windows users

moboqe:
> @moboqe this is not merged, It works if you dont have space in your path, I made it so it can bypass the issue created by this PR. I am not sure if this PR works for other platforms. @cosmtrek I think the best is to revert this PR since the author is not responding and it is affecting all windows users

I mean this doesn't work for me related for **this** issue.
Yep, I also fixed those issue by adding check for space in path

PR #653, opened 2024-09-22
Refactor `startCmd()` related code

8LWXpg:
This PR does the following:
- Removed the return of `Stdout`, `Stderr` in `startCmd()`.
- Removed code related to returned `stderr` and `stdout`.
- The command now maps `Stdin`, allowing it to support input from stdin..
- Removed the check for the `.exe` file extension on Windows, since executing `.exe` files without the extension is supported.


runner/util_linux.go
```diff
@@ -28,8 +29,15 @@ func (e *Engine) killCmd(cmd *exec.Cmd) (pid int, err error) {
 	return
 }
 
-func (e *Engine) startCmd(cmd string) (*exec.Cmd, io.ReadCloser, io.ReadCloser, error) {
-	c := exec.Command("/bin/sh", "-c", cmd)
+func (e *Engine) startCmd(cmd string) (c *exec.Cmd, err error) {
+	c = exec.Command("/bin/sh", "-c", cmd)
 	f, err := pty.Start(c)
```
8LWXpg, line 34: It appears that `f.Close()` must be called before exit or the process will be stuck.

Issue #640, opened 2024-09-04
Win 10 Reload fails unable to delete main.exe


coraxwolf:
I am running on Windows 10 using git bash (have tried in power shell as well). When I make a change air attempts to reload and fails to end the current running task which causes an Access Denied error when build tries to replace the current main.exe.  This results in an endless looping of trying to kill the process and starting the app causing address already in use errors which repeat endlessly until I cancel out of Air.

Here is an excerpt of the air debug logs:
```
[15:27:46] running...
[15:27:46] running process pid 29488
{"time":"2024-09-04T15:27:46.9243759-05:00","level":"INFO","source":{"function":"coraxwolf/contacts-go/cmd/server.(*Server).Start","file":"C:/Users/j*****/Documents/Projects/golang/Contacts/cmd/server/main.go","line":28},"msg":"Server is starting on :3000"}
[15:28:00] event: WRITE         "C:\\Users\\*****\\Documents\\Projects\\golang\\Contacts\\main.go"
[15:28:00] main.go has changed
[15:28:00] event: WRITE         "C:\\Users\\******\\Documents\\Projects\\golang\\Contacts\\main.go"
[15:28:00] main.go has changed
[15:28:01] flushing events
[15:28:01] main.go has changed
[15:28:01] building...
[15:28:01] trying to kill pid 29488, cmd [cmd /c C:\Users\*****\Documents\Projects\golang\Contacts\tmp\main.exe]
[15:28:01] failed to kill PID 29488, error: exit status 1
[15:28:01] failed to remove C:\Users\*******\Documents\Projects\golang\Contacts\tmp\main.exe, error: remove C:\Users\*********\Documents\Projects\golang\Contacts\tmp\main.exe: Access is denied.
```

I am sure this is something in my settings and/or how I am trying to use this tool.

xiantang:
i think it should be fixed after #670 got merged, pls feel free to reopen the issue

Issue #492, opened 2023-10-31
send_interrupt not working on Linux


osechet:
Hi!
I'm trying to use the `send_interrupt` option on Linux but cannot make it work. Here is my config file:
```
root = "."
tmp_dir = "tmp"

[build]
  args_bin = []
  bin = "tmp/foobar"
  cmd = "go build -o ./tmp/foobar ."
  delay = 0
  exclude_dir = ["assets", "tmp", "vendor"]
  exclude_file = []
  exclude_regex = ["_test\\.go"]
  exclude_unchanged = false
  follow_symlink = false
  full_bin = ""
  include_dir = []
  include_ext = ["go", "tpl", "tmpl", "html"]
  include_file = []
  kill_delay = "10s"
  log = "build-errors.log"
  poll = false
  poll_interval = 0
  rerun = false
  rerun_delay = 500
  send_interrupt = true
  stop_on_error = false

[color]
  app = ""
  build = "yellow"
  main = "magenta"
  runner = "green"
  watcher = "cyan"

[log]
  main_only = false
  time = false

[misc]
  clean_on_exit = false

[screen]
  clear_on_rebuild = false
  keep_scroll = true
```

When running in debug mode, here are the logs I get after pressing _Ctrl+C_:
```
{"time":"2023-10-31T17:37:36.116222472+01:00","level":"INFO","prefix":"main","file":"server.go","line":"84","message":"Starting to listen on :8443..."}
exit in start
cleaning...
wating for      close watchers..
waiting for buildRun...
trying to kill pid 28390, cmd [/bin/sh -c /workspace/tmp/foobar]
waiting for clean ...
waiting for exit...
exited
see you again~
```

I'm expecting a log from my server saying it's gracefully stopping (working fine without air) that never appears.

I checked the code of [engine.go](https://github.com/cosmtrek/air/blob/eb5b1b501bf2a69316cfc75648c2eebba98d0553/runner/engine.go#L503C4-L503C4). It looks like I should also get a "cmd killed, pid: %d" log. But it isn't the case, as if the air process was closed before it reached that line. 

Is there something wrong with my config file?

joe-xonasystems:
i've hit the same problem. works fine in [CompileDaemon](https://github.com/githubnemo/CompileDaemon) (simpler tool which may not cover the same use cases as Air)

joe-xonasystems:
> as if the air process was closed before it reached that line.

[...]

Issue #334, opened 2022-10-10
Gin Golang server seems to hang after 3 or 4 requests while using Air


hgoona:
I'm new to dev so Im hoping this isn't a installation problem but I'm noticing that my Go server stops responding after 3 or sometimes 4 responses.

I'm on windows, latest version of Go.

I'm building an api on localhost and sending requests through Postman (and double checked the lack of response with Hoppscotch as well). In both situations I noticed the responses from my server stop after 3 or 4 requests; where I then need to Ctrl+C my vscode file, and re-run `air`.

However, if I use `go run .` on my code instead of air, i note that my server correctly responds without hanging.
Do I have an installation or configuration problem to look for, or is this a bug in Air?

andradei:
@hgoona it's been almost a year. Sometimes `air` hangs in the air :smile: for me too and `Ctrl + C` doesn't stop it.
I have to manually kill it from another terminal with something like `kill -9 <pid of air process>`.

But after a few instances of this, it "magically" starts working again.

rowbawts:
I just started using Air in the past 2 days and I'm having this issue as well and it's happening often enough that its very annoying. Are there any workarounds for this?

duizabojul:
Same issue on mac docker with orbstack, can't ctrl C to stop air process

dimmerz92:
I also occasionally get this issue.
If I kill the terminal and restart air, the port is usually still occupied, so I am required to kill the process `kill -9 <PID>` before starting air again.
Otherwise, loving air.

andradei:
I just learned <kbd>CTRL + \\</kbd> kills air (it causes a ton of goroutine panics) and frees the used port.
Still not a solution.

Issue #131, opened 2020-12-14
Means of manually triggering restart?


bensaufley:
First off: thanks for this! It's immensely useful to have a fast and full-featured Go reloader like this. I came here from Fresh and I appreciate the extra functionality this package offers on top of that.

I've noticed a few issues others have pointed out in my short time using Air, like #129, and I think one way to mitigate the frustration factor here while these sorts of things are worked out would be to enable a manual restart trigger [à la Nodemon's `rs`](https://github.com/remy/nodemon#manual-restarting). Is this existing and undocumented, or have I missed documentation, or is this something that could be added?

The other way I've generally handled this problem in other codebases is just saving a file with no changes, but Air currently explicitly ignores those events—could that be disabled or configured?

bensaufley:
I actually just found the `exclude_unchanged` flag, so that answers the second question.

Right now it's failing to reload more times than not, either because it says a file changed but then does nothing, or because it reloads and the server port is still taken from a previous task (#129 possibly?).

PR #670, opened 2024-11-01
wait for killfunc completion when shutting down current app

### LINKED ISSUE #669: Wait for complete current app shutdown before starting a new build run
When air is used with applications that take some time to shut down, it can become quite frequent to see problems with file access, port availability and even lockups that require ctrl-c or worse to get out of. There has been a few releases with fixes in this area, but the problems sadly persist and have a negative impact on development flow.

Air should make sure that the currently running application has been able to perform any required cleanup, and has been completely shutdown, before kicking off a new build.

istyf:
This PR attempts to fix #669 by reworking how the killFunc is signalled that a shutdown has been requested.

Instead of reading directly from e.binStopCh (that creates a data race with the writer) it closes on a returned shutdown channel that is used from a new mutexed e.stopBin() func. The shutdown channel takes a channel of int, where the int is the "exit code" passed back to the requestor. It isn't actually the exit code of the process, but a code that can be (and currently is) used to os.Exit the entire process if shutdown failed.

The atomic.Bool change is unrelated to the actual issue, but was required to be able to run the unit tests with -race to detect if there were any races between all the channels and goroutine handling going on in this code.

Reading through the issue list there is a good possibility that this PR will fix, or considerably reduce, the problems seen in #640 #492 #334 and do away with the need for #131 .

This PR has been unit tested and development tested on wsl2 linux and an Apple M2 MBP.

### DIFF
--- a/runner/engine.go
+++ b/runner/engine.go
@@ -29,8 +29,10 @@ type Engine struct {
 	watcherStopCh  chan bool
 	buildRunCh     chan bool
 	buildRunStopCh chan bool
-	binStopCh      chan<- chan int
-	exitCh         chan bool
+	// binStopCh is a channel for process termination control
+	// Type chan<- chan int indicates it's a send-only channel that transmits another channel(chan int)
+	binStopCh chan<- chan int
+	exitCh    chan bool
 
 	mu            sync.RWMutex
 	watchers      uint
@@ -605,11 +607,9 @@ func (e *Engine) stopBin() {
 	e.mainDebug("shutdown completed in %v", time.Since(start))
 
 	exitCode := make(chan int)
-	wasRunning := false
 
 	e.withLock(func() {
 		if e.binStopCh != nil {
-			wasRunning = true
 			e.mainDebug("sending shutdown command to killfunc")
 			e.binStopCh <- exitCode
 			e.binStopCh = nil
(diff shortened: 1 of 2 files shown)


xiantang:
i will take look into this pr. lets see how to make the logic easy to understand

istyf:
@xiantang Great! Let me know if I can be of any help. I suppose I could add a few comments here and there to make the intention more clear.

xiantang:
Very Bidirectional Channels Design

runner/engine.go
```diff
@@ -583,9 +601,31 @@ func (e *Engine) runBin() error {
...
+		if e.binStopCh != nil {
+			e.mainDebug("sending shutdown command to killfunc")
+			e.binStopCh <- exitCode
+			e.binStopCh = nil
+		} else {
+			close(exitCode)
+		}
+	})
+	if ret := <-exitCode; ret != 0 {
+		os.Exit(ret)
```
ccoVeille, line 621: Your refactoring is great, I love it.

I would have appreciated if this could be tested.

One solution for this is to define an exiter in the engine.

The exiter could be assigned to os.Exit by default, and overloaded via the tests to return a t.Fatal if the code is not the one expected

What do you think about it?

xiantang: looks good i will raise a pr for that.

ccoVeille: Thanks

xiantang: https://github.com/air-verse/air/pull/729

PR #672, opened 2024-11-01
do not sleep to kill a process that has already cleaned up and exited by itself

### LINKED ISSUE #671: Do not sleep the full kill_delay when using send_interrupt and the current application shuts down by itself
Currently when send_interrupt is configured to true (for linux and unix only), the killCmd function sleeps for kill_delay before following up with a kill command as well.

If the target process honours the initial interrupt, cleans up and exits by itself, there is no need to sleep for a delay to try to kill a process that is no longer there. Air should instead wake up and continue as soon as the process has closed down, and only go on to send a kill to processes that ignore the interrupt or take too long to shut down.

istyf:
This PR fixes #671 by replacing the time.Sleep in killCmd (for linux and unix/mac) with a cancelable context that can be used to stop the kill timer and return to the caller as soon as the interrupted process has closed down.

While developing this PR I took the liberty of deleting the unit test for testing killing of "non existing" processes for reasons outlined in 52428bf7f43a18e0e96774071cccbf761d8bb1ca.

This PR has been tested on wsl2 linux and an Apple M2 MBP.

### DIFF
--- a/runner/util_linux.go
+++ b/runner/util_linux.go
@@ -4,8 +4,8 @@ import (
 	"context"
 	"errors"
 	"io"
-	"os"
 	"os/exec"
+	"syscall"
 	"time"
 
 	"github.com/creack/pty"
@@ -24,8 +24,8 @@ func (e *Engine) killCmd(cmd *exec.Cmd) (pid int, err error) {
 
 	if e.config.Build.SendInterrupt {
 		e.mainDebug("sending interrupt to process %d", pid)
-		// Sending a signal to make it clear to the process that it is time to turn off
-		if err = cmd.Process.Signal(os.Interrupt); err != nil {
+		// Sending a signal to make it clear to the process group that it is time to turn off
+		if err = syscall.Kill(-pid, syscall.SIGINT); err != nil {
 			return
 		}
 		// the kill delay is 0 by default unless the user has configured send_interrupt=true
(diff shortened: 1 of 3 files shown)


istyf:
@xiantang What is the status/probability of accepting this PR?

xiantang:
could u rebase to master?

xiantang:
Because I'm doing https://github.com/air-verse/air/pull/809, I think there may be conflicts.

xiantang:
I need to merge #809 to master first, because once I remove the pty, Linux behavior will be broken, and it cannot clean child processes correctly.

istyf:
> I need to merge #809 to master first, because once I remove the pty, Linux behavior will be broken, and it cannot clean child processes correctly.

I do not understand. Why would you merge something that breaks Linux behaviour?

xiantang:
Oh, I see, you just changed the killcmd code, it should be okay. because for linux is still using pty to start a new process

runner/util_linux.go
```diff
@@ -12,20 +14,65 @@ import (
...
+			return
+		}
+	}()
+
+	results := make([]error, 0, 2)
 
-	// Wait releases any resources associated with the Process.
-	_, _ = cmd.Process.Wait()
-	return
+	for {
```
xiantang: Could you make this block neat? After that, I will merge this PR.

istyf: I added comments to clarify the code and even added a >= 2 instead of == 2 to please Copilot. The fix it proposes above does not even compile and the fix it proposes here would actually introduce a bug rather than fix one.

xiantang:
Hello @istyf, could you help resolve the conflict?

istyf:
Upstream merged into PR and conflict resolved. However this has now not been tested on my side.

xiantang:
Okay, please test this feature again as well, thanks.

Issue #673, opened 2024-11-02
Copying of stdout/stderr causes frequent deadlocks under WSL2


istyf:
Running the unit tests have a near (if not) 100% reproducible deadlock between all of the different output writers, caused by contention of a semaphore deep in the posix file system to be able to output text.

I have not had the time to root out the exact reason, but the use of io.Copy(os.Stdout, stdout) seems to be the culprit, because if that is changed the deadlock is no longer reproducible.

It seems reasonable that this deadlock could also happen within air itself and not only in the unit tests.

PR #674, opened 2024-11-02
replace io.Copy with a scanner to avoid output related deadlocks on WSL2

istyf:
This PR is a fix for #673 

I have no clear idea what the actual problem is here, but deadlocks are very frequent when running the unit tests under WSL2 and seem to be related to the use of io.Copy to copy stdout and stderr

A deadlock occurs when multiple goroutines get parked indefinitely waiting for an rwlock deep down in the posix file system to be able to write to stdout...

Before merging, please check the change in runCommand. Is there a reason why this copying is synchronous (all of stdout before starting to copy stderr) or should they be asynchronous coroutines just as in runBin?

### DIFF
--- a/runner/engine.go
+++ b/runner/engine.go
@@ -435,8 +435,8 @@ func (e *Engine) runCommand(command string) error {
 		stderr.Close()
 	}()
 
-	go copyOutput(os.Stdout, stdout)
-	go copyOutput(os.Stderr, stderr)
+	copyOutput(os.Stdout, stdout)
+	copyOutput(os.Stderr, stderr)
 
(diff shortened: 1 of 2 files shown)


istyf:
@xiantang Sorry. It took me 6+ months to notice your request... The conflict is now resolved.

xiantang:
Could you help fix some lints? I noticed that the linting failed the pipeline.

xiantang:
I would like to merge this PR first after you fix the lint.

istyf:
Delinted.

xiantang:
thank merged

Issue #655, opened 2024-09-25
Sometimes `air` panics when ran


MilkeeyCat:
Sometimes when I run command `air` I get such an error
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x90 pc=0x527620]

goroutine 40 [running]:
regexp.(*Regexp).doExecute(0xc00032be00?, {0x0?, 0x0?}, {0xc000380050?, 0x7fcd05?, 0xc00032be26?}, {0x0?, 0xc00032be00?}, 0x4d?, 0x0, ...)
        /home/alex/Downloads/go/src/regexp/exec.go:527 +0x80
regexp.(*Regexp).doMatch(...)
        /home/alex/Downloads/go/src/regexp/exec.go:514
regexp.(*Regexp).Match(...)
        /home/alex/Downloads/go/src/regexp/regexp.go:537
github.com/cosmtrek/air/runner.(*Engine).isExcludeRegex(0xc0001d6000?, {0xc00032be00, 0x4d})
        /home/alex/go/pkg/mod/github.com/cosmtrek/air@v1.52.0/runner/util.go:149 +0xcb
github.com/cosmtrek/air/runner.(*Engine).watchPath.func1()
        /home/alex/go/pkg/mod/github.com/cosmtrek/air@v1.52.0/runner/engine.go:257 +0x3d3
created by github.com/cosmtrek/air/runner.(*Engine).watchPath in goroutine 1
        /home/alex/go/pkg/mod/github.com/cosmtrek/air@v1.52.0/runner/engine.go:224 +0x23f
```

I had version 1.52.0 but the place where the error occurred doesn't seem to be updated since `1.52.0` version

OS: Linux

cosmtrek:
@MilkeeyCat please provide more details to reproduce it

MilkeeyCat:
I don't have a way how to reproduce it, it just happens sometimes

istyf:
This seems to be the same issue as #530 that is reported on an even older version of air. It seems that the compiler regex is nil when it is dereferenced in regexp/exec.go:527, but it isn't immediately obvious what could have caused that when reading through the code higher up in the call stack ... This smells like some sort of memory issue ...

xiantang:
should be fixed by https://github.com/air-verse/air/pull/677

Issue #530, opened 2024-02-16
Bug report: one-off `panic: runtime error: invalid memory address or nil pointer dereference` at startup


ChrysmOre:
Have recently integrated Air into a repo at work, everything has been running fine until today on a fresh pull from the repository, everything in the same state as has been working perfectly but have hit a panic:

```
building...
***
2024-02-16 14:08:32 panic: runtime error: invalid memory address or nil pointer dereference
2024-02-16 14:08:32 [signal SIGSEGV: segmentation violation code=0x1 addr=0x90 pc=0x518949]
2024-02-16 14:08:32 
2024-02-16 14:08:32 goroutine 37 [running]:
2024-02-16 14:08:32 regexp.(*Regexp).doExecute(0x28?, {0x0?, 0x0?}, {0xc000126750?, 0xc0001266f5?, 0x4?}, {0x0?, 0x2d?}, 0x28?, 0x0, ...)
2024-02-16 14:08:32     /usr/local/go/src/regexp/exec.go:527 +0x89
2024-02-16 14:08:32 regexp.(*Regexp).doMatch(...)
2024-02-16 14:08:32     /usr/local/go/src/regexp/exec.go:514
2024-02-16 14:08:32 regexp.(*Regexp).Match(...)
2024-02-16 14:08:32     /usr/local/go/src/regexp/regexp.go:539
2024-02-16 14:08:32 github.com/cosmtrek/air/runner.(*Engine).isExcludeRegex(0xc000246000?, {0xc0001266f0, 0x2d})
2024-02-16 14:08:32     /go/pkg/mod/github.com/cosmtrek/air@v1.49.0/runner/util.go:149 +0xcb
2024-02-16 14:08:32 github.com/cosmtrek/air/runner.(*Engine).watchPath.func1()
2024-02-16 14:08:32     /go/pkg/mod/github.com/cosmtrek/air@v1.49.0/runner/engine.go:255 +0x3c9
2024-02-16 14:08:32 created by github.com/cosmtrek/air/runner.(*Engine).watchPath in goroutine 1
2024-02-16 14:08:32     /go/pkg/mod/github.com/cosmtrek/air@v1.49.0/runner/engine.go:222 +0x24e
2024-02-16 14:08:32 panic: runtime error: invalid memory address or nil pointer dereference
2024-02-16 14:08:32 [signal SIGSEGV: segmentation violation code=0x1 addr=0x90 pc=0x518949]
2024-02-16 14:08:32 
[...]

PR #677, opened 2024-11-02
remove a data race by compiling the exclude regexes in the preprocess phase instead

### LINKED ISSUE #668: Segfault When `air` Command Ran
I got a segfault when I try to run the command `air` but when I reset my PC and run `air` again it refreshes my page but I instantly get segfault
```
watching public
!exclude tmp
watching views
watching views/auth
watching views/components
watching views/css
watching views/home
watching views/layouts
building...
running...
building static files for development
2024/10/19 23:20:29 INFO HTTPS Server Started listenAddr=:3000
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x90 pc=0x5ec12a0]

goroutine 40 [running]:
regexp.(*Regexp).doExecute(0xc0000b2780?, {0x0?, 0x0?}, {0xc00024a120?, 0x5d?, 0x60?}, {0x0?, 0x0?}, 0x0?, 0x0, ...)
        /usr/local/go/src/regexp/exec.go:527 +0x80
regexp.(*Regexp).doMatch(...)
        /usr/local/go/src/regexp/exec.go:514
regexp.(*Regexp).Match(...)
        /usr/local/go/src/regexp/regexp.go:533
github.com/air-verse/air/runner.(*Engine).isExcludeRegex(0xc000222000?, {0xc0000b2780, 0x5d})
        /Users/lidldev/go/pkg/mod/github.com/air-verse/air@v1.61.0/runner/util.go:167 +0xcb
github.com/air-verse/air/runner.(*Engine).watchPath.func1()
        /Users/lidldev/go/pkg/mod/github.com/air-verse/air@v1.61.0/runner/engine.go:256 +0x3ec
created by github.com/air-verse/air/runner.(*Engine).watchPath in goroutine 1
        /Users/lidldev/go/pkg/mod/github.com/air-verse/air@v1.61.0/runner/engine.go:223 +0x23f
```

istyf:
This PR should fix #668 #655 and #530

The regular expressions, that are used to determine if files should be excluded, are compiled lazily on first request. This causes different kinds of timing related problems, including panics because multiple goroutines will end up compiling the expressions, while other goroutines may return with data that isn't complete.

### DIFF
--- a/runner/config.go
+++ b/runner/config.go
@@ -62,16 +62,6 @@ type cfgBuild struct {
 }
 
 func (c *cfgBuild) RegexCompiled() ([]*regexp.Regexp, error) {
-	if len(c.ExcludeRegex) > 0 && len(c.regexCompiled) == 0 {
-		c.regexCompiled = make([]*regexp.Regexp, 0, len(c.ExcludeRegex))
-		for _, s := range c.ExcludeRegex {
-			re, err := regexp.Compile(s)
-			if err != nil {
-				return nil, err
-			}
-			c.regexCompiled = append(c.regexCompiled, re)
-		}
-	}
 	return c.regexCompiled, nil
 }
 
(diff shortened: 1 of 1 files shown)


xiantang:
look good, we can move this logic to startup

xiantang (approved):

PR #725, opened 2025-01-11
Feature: Show build errors when using proxy

Polo123456789:
This is a feature that I missed when coming from the Javascript wold.

And, being relatively new to Go, having the build fail and testing the fronted with and old version of the program made me lose a good amount of time before I noticed.

Being new to open source and Go I'm sure that this PR will need more work, so I'll appreciate any feedback.

Thanks in advance.

Example:
image

### DIFF
--- a/runner/proxy_stream.go
+++ b/runner/proxy_stream.go
@@ -1,6 +1,7 @@
 package runner
 
 import (
+	"encoding/json"
 	"sync"
 	"sync/atomic"
 )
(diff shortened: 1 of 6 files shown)


runner/proxy_stream.go
```diff
@@ -42,13 +61,36 @@ func (stream *ProxyStream) RemoveSubscriber(id int32) {
...
+}
+
+func (m StreamMessage) AsSSE() string {
+	s := "event: " + string(m.Type) + "\n"
+	s += "data: " + stringify(m.Data) + "\n"
+	return s + "\n"
+}
+
+func stringify(v any) string {
+	b, _ := json.Marshal(v)
```
xiantang, line 94: if err != nil {
        return fmt.Sprintf("{\"error\":\"Failed to marshal message: %s\"}", err)
    }

xiantang:
Pls add tests for build failure scenarios:

PR #727, opened 2025-01-12
Avoid duplicating cors header.

### LINKED ISSUE #726: Proxy unconditionally adds `Access-Control-Allow-Origin`
This is the problematic code:

```
func (p *Proxy) proxyHandler(w http.ResponseWriter, r *http.Request) {
	w.Header().Set("Access-Control-Allow-Origin", "*")
```

If the API behind the proxy also adds CORS headers, you get some final response with duplicate headers:

```
Access-Control-Allow-Origin: *
Access-Control-Allow-Origin: *
```

This is _usually_ not a problem, but some clients are less resilient (or "more strict") than others and choke on the redundant headers.

Ideally, the proxy should not add this header if it's already present.

brandonbloom:
Fixes #726

### DIFF
--- a/runner/proxy.go
+++ b/runner/proxy.go
@@ -71,8 +71,6 @@ func (p *Proxy) injectLiveReload(resp *http.Response) (string, error) {
 }
 
 func (p *Proxy) proxyHandler(w http.ResponseWriter, r *http.Request) {
-	w.Header().Set("Access-Control-Allow-Origin", "*")
-
 	appURL := r.URL
 	appURL.Scheme = "http"
 	appURL.Host = fmt.Sprintf("localhost:%d", p.config.AppPort)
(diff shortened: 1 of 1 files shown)


runner/proxy.go
```diff
@@ -131,6 +129,7 @@ func (p *Proxy) proxyHandler(w http.ResponseWriter, r *http.Request) {
 			w.Header().Add(k, v)
 		}
 	}
+	w.Header().Set("Access-Control-Allow-Origin", "*")
```
xiantang, line 132: // Only set CORS header if it's not already present in the response
	if resp.Header.Get("Access-Control-Allow-Origin") == "" {
		w.Header().Set("Access-Control-Allow-Origin", "*")
	}

xiantang (approved):

PR #740, opened 2025-02-16
chore: bump go to 1.24

linzhengen:
- Upgraded Go 1.23 to Go 1.24.
- Updated GitHub workflows (build, release, smoke tests) to use Go 1.24 with go.mod file.
- Changed Dockerfile base images from golang:1.23 to golang:1.24.
- Migrated golangci-lint to v2 in the Makefile and CI.
  - doc: https://golangci-lint.run/product/migration-guide/
- Revised README files (EN, zh_cn, zh_tw, ja) to reflect the new Go version.
- Updated go.mod and adjusted dependency versions for compatibility.


MoritzKrafeld:
@cosmtrek do you mind taking a look at this? We would like to use this :smile:

MoritzKrafeld:
Hey @xiantang would you mind taking a look at this? :smile:

im7mortal:
@linzhengen  I feel like this PR should be split into smaller ones.

You mentioned that you're bumping the version to 1.24, but you've also included many smaller changes that should be verified separately—especially the major upgrade of golangci-lint.

I suggest keeping only the changes strictly required for the Go 1.24 bump in this PR, and creating separate issues for each of your other suggestions.

linzhengen:
@im7mortal 
Thanks for the feedback!

You're absolutely right that this PR currently includes multiple changes, and I agree it would be clearer if it were split up.

Just to provide some context: I ended up upgrading `golangci-lint` because the version we were using (`v1.61.0`) failed with the following error after bumping to Go 1.24:

```
Error: can't load config: the Go language version (go1.23) used to build golangci-lint is lower than the targeted Go version (1.24)
Failed executing command with error: can't load config: the Go language version (go1.23) used to build golangci-lint is lower than the targeted Go version (1.24)
```

Since continuing with `v1.61.0` wasn't feasible, I upgraded to the latest available version. I did consider sticking with the latest `v1.x`, but decided to try `v2` for now.

MoritzKrafeld:
I took your PR and split it up into multiple ones so that we can update to go 1.24 :smile:  
Here is the PR for golangci-lint to be compatible with go 1.24: https://github.com/air-verse/air/pull/762  
And this is the PR which then updates go to 1.24: https://github.com/air-verse/air/pull/763 
I'm currently working on a PR to also update golangci-lint to 2.x once we are on go 1.24 but this needs a bit more time :smile:

PR #742, opened 2025-02-19
Resolve root directory if symlinked

### LINKED ISSUE #531: air not working when project directory is a symlink
Hi, it seems that air is unable to watch files when accessing the project through a symlink. Here are the steps to reproduce the problem:

    $ mkdir project
    $ cd project
    $ cat > main.go <<EOF
    package main
    func main() {
    	for {}
    }
    EOF
    $ go mod init project
    $ air
      __    _   ___  
     / /\  | | | |_) 
    /_/--\ |_| |_| \_ v1.49.0, built with Go go1.21.6
    
    mkdir /tmp/tmp.nE9SAKeGRq/project/tmp
    watching .
    !exclude tmp
    building...
    running...
    ^Ccleaning...
    see you again~
    $ cd ..
    $ ln -s project symlink
    $ air
      __    _   ___  
     / /\  | | | |_) 
    /_/--\ |_| |_| \_ v1.49.0, built with Go go1.21.6
    
    mkdir /tmp/tmp.nE9SAKeGRq/symlink/tmp
    building...
    running...
    ^Ccleaning...
    see you again~

As you can see, when `cd`ing into the `symlink` file, air doesn't print the `watching .` message, and indeed doesn't watch for file updates.

How can I solve this issue? Thanks!

caleb-fringer:
While the follow_symlink option allows for dereferencing symlinked directories in the project, it does not resolve the root directory to a real directory if it is symbolically linked. This becomes an issue if you are accessing a project directory from a symlink, as filepath.Walk ignores symbolic links.

By adding an additional de-referencing step to Config.preprocess(), this PR resolves Config.Root directories that are themselves symbolic links to their target path so that Engine.watching() will correctly walk the root path.

Closes #531

### DIFF
--- a/runner/config.go
+++ b/runner/config.go
@@ -560,7 +560,7 @@ func (c *Config) preprocess(args map[string]TomlInfo) error {
 		if err = os.Chdir(cwd); err != nil {
 			return err
 		}
-		c.Root = cwd
+		c.Root = "."
 	}
 	c.Root, err = expandPath(c.Root)
 	if err != nil {
--- a/runner/config.go
+++ b/runner/config.go
@@ -560,7 +560,7 @@ func (c *Config) preprocess(args map[string]TomlInfo) error {
 		if err = os.Chdir(cwd); err != nil {
 			return err
 		}
-		c.Root = "."
+		c.Root = cwd
 	}
 	c.Root, err = expandPath(c.Root)
 	if err != nil {
--- a/runner/util.go
+++ b/runner/util.go
@@ -271,6 +271,7 @@ func expandPath(path string) (string, error) {
 
 // Dereferences a symbolic link to its relative path.
 // If the path is not a symlink, simply returns the path.
+// If the file does not exist, return the path and no err.
 func derefLink(path string) (string, error) {
 	ok, err := isSymlink(path)
 	if err != nil {
--- a/runner/util.go
+++ b/runner/util.go
@@ -276,20 +276,17 @@ func expandPath(path string) (string, error) {
 
 	expanded, err := filepath.Abs(expanded)
 	if err != nil {
-		return "", fmt.Errorf("Error getting absolute path to %v: %v", expanded, err)
+		return "", fmt.Errorf("error getting absolute path to %v: %w", expanded, err)
 	}
 
 	// filepath.EvalSymlinks only works on real files
 	dereferenced, err := filepath.EvalSymlinks(expanded)
-	if err != nil {
-		// The error is something more serious than the path not existing, fail.
-		if !errors.Is(err, fs.ErrNotExist) {
-			return "", fmt.Errorf("Unexpected error while dereferencing %v: %v", expanded, err)
-		}
-	} else {
-		// The error is simply that the path is not a real file.
-		// Continue without dereferencing.
-		expanded = dereferenced
+	if err == nil {
+		return dereferenced, nil
+	}
+
+	if !errors.Is(err, fs.ErrNotExist) {
+		return "", fmt.Errorf("unexpected error while dereferencing %v: %w", expanded, err)
 	}
 
 	return expanded, nil
--- a/runner/util.go
+++ b/runner/util.go
@@ -274,6 +274,9 @@ func expandPath(path string) (string, error) {
 func derefLink(path string) (string, error) {
 	ok, err := isSymlink(path)
 	if err != nil {
+		if os.IsNotExist(err) {
+			return path, nil
+		}
 		return "", err
 	}
 	if !ok {
--- a/runner/util.go
+++ b/runner/util.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"log"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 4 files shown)


caleb-fringer:
Tested on 5.15.167.4-microsoft-standard-WSL2 running Ubuntu 22.04.5 LTS.

xiantang:
can i have any unit test coverage?

caleb-fringer:
After adding unit tests for the isSymlink function and expandPath, I am observing perplexing behavior with some of the runner unit tests. 

If I cd into `runner/` and run `go test . -v`, then the following unit tests fail:
image

However, if run individually, all except TestCheckRunEnv passes:

image

I suspect this is due to some discrepancy between how the current path is set when unit testing a module versus individual tests, or perhaps some state issue with TestCheckRunEnv. I could use some help understanding this test in particular.

Attached is the output of: `go test . -v`, which contains logs for the failing tests:
[failing-tests.txt](https://github.com/user-attachments/files/19276058/failing-tests.txt)

xiantang:
lets see if its work in CI

caleb-fringer:
I was able to fix the unit tests I broke by adding an explicit check to see if the path passed to `derefLink` has been deleted. However, this function is only ever called on `config.Root`, and in the failing test cases (for instance, `TestNewEngine`) this was set to `toml_root`. This path that does not exist while running the whole test suite. I assumed that this path must already exist by time the test is run, which is what caused Lstat to return an error.

Furthermore, I added a unit test for calling `expandPath` on relative paths, and I am observing the same issue as before: when I use `go test .` it fails at `os.Getwd()` because the path no longer exists. This only occurs when running test on the package, not when individually running the test with `go run test . -run TestExpandPathWithRelPath`

It really seems like some of the test cases are modifying the runtime environment that other tests depend upon. Can you provide any insight to how the test environment is constructed and modified by other tests?

xiantang:
@caleb-fringer hello, do u still have time to work on this feature? i have enough time to review code now!

xiantang:
and i already make my unit tests more stable now, could u rebase to master and lets run the UT and see whats happen

xiantang:
this is my fault, in flag_test.go i didnt restore to the old cmd with t.Cleanup, so could u help me do this and btw use filepath.Join, not path.Join, for file paths.
after that i think the tests are good

caleb-fringer:
I realized I was duplicating some logic from the filepaths stdlib. I refactored the `expandPath()` function to simply handle expanding the path directly using `filepath.Abs()` and `filepath.ExpandSymlinks()`. Wrote a regression test and ran it on my system (PopOS 24.04). All unit tests are now passing.

Let me know if you find anything else that needs attention.

xiantang:
could u help me fix the golang-lint?

runner/util.go
```diff
@@ -258,23 +259,40 @@ func copyOutput(dst io.Writer, src io.Reader) {
...
-	var err error
-	wd, err := os.Getwd()
+
+	expanded, err := filepath.Abs(expanded)
 	if err != nil {
-		return "", err
-	}
-	if path == "." {
-		return wd, nil
+		return "", fmt.Errorf("Error getting absolute path to %v: %v", expanded, err)
```
xiantang: pls help fix this staticcheck lint?

runner/util.go
```diff
@@ -258,23 +259,40 @@ func copyOutput(dst io.Writer, src io.Reader) {
...
 	}
-	if strings.HasPrefix(path, "./") {
-		return wd + path[1:], nil
+
+	// filepath.EvalSymlinks only works on real files
+	dereferenced, err := filepath.EvalSymlinks(expanded)
+	if err != nil {
+		// The error is something more serious than the path not existing, fail.
+		if !errors.Is(err, fs.ErrNotExist) {
+			return "", fmt.Errorf("Unexpected error while dereferencing %v: %v", expanded, err)
```
xiantang: ditto

runner/util.go
```diff
@@ -258,23 +259,40 @@ func copyOutput(dst io.Writer, src io.Reader) {
...
+		return "", fmt.Errorf("Error getting absolute path to %v: %v", expanded, err)
 	}
-	if strings.HasPrefix(path, "./") {
-		return wd + path[1:], nil
+
+	// filepath.EvalSymlinks only works on real files
+	dereferenced, err := filepath.EvalSymlinks(expanded)
+	if err != nil {
+		// The error is something more serious than the path not existing, fail.
+		if !errors.Is(err, fs.ErrNotExist) {
```
xiantang: Do we need to have `else` case for this trunk

runner/util.go
```diff
@@ -258,23 +259,40 @@ func copyOutput(dst io.Writer, src io.Reader) {
...
+			return "", fmt.Errorf("Unexpected error while dereferencing %v: %v", expanded, err)
+		}
+	} else {
+		// The error is simply that the path is not a real file.
```
xiantang: the comment looks wrong

runner/util.go
```diff
@@ -258,23 +259,40 @@ func copyOutput(dst io.Writer, src io.Reader) {
...
-	if path == "." {
-		return wd, nil
+		return "", fmt.Errorf("Error getting absolute path to %v: %v", expanded, err)
 	}
-	if strings.HasPrefix(path, "./") {
-		return wd + path[1:], nil
+
+	// filepath.EvalSymlinks only works on real files
+	dereferenced, err := filepath.EvalSymlinks(expanded)
+	if err != nil {
```
xiantang: this will be more clear

```go
 dereferenced, err := filepath.EvalSymlinks(expanded)
 if err == nil {
     return dereferenced, nil
 }

 if !errors.Is(err, fs.ErrNotExist) {
     return "", fmt.Errorf("unexpected error while dereferencing %v: %w", expanded,
 err)
 }

 return expanded, nil
```

caleb-fringer: Fixed in baeb8f8. I think I originally was doing the de-referencing step before resolving to an absolute path, which made the logic more tricky. Didn't clean it up properly when I refactored.

runner/config.go
```diff
@@ -560,7 +560,7 @@ func (c *Config) preprocess(args map[string]TomlInfo) error {
 		if err = os.Chdir(cwd); err != nil {
 			return err
 		}
-		c.Root = "."
+		c.Root = cwd
```
caleb-fringer, line 563: I think I changed this in my first attempt at implementing this feature in my PR last year. However, after simplifying the process to simply use expandPath to dereference and expand the root, this no longer seems necessary.

xiantang:
@caleb-fringer i used claude code to fix these failed issues, and now the CI runs great now

xiantang:
It looks good to me now; let me merge it. Thanks

xiantang (approved):

PR #743, opened 2025-02-20
Add JSON Schema for Configuration File

arhea:
IDE Extensions like [Even Better TOML](https://github.com/tamasfe/taplo) in VSCode use [JSON Schema](https://json-schema.org/) for validating TOML configurations. This pull request adds a JSON Schema for the `.air.toml` configuration file that can be placed in the root of projects.

This file can then be consumed by these extensions from the GitHub project.


xiantang:
thank you, i wanna know, if we add more options, we need to change this file as well, right?

rowan-gud:
Ah... I should've checked this repo first... I opened a PR to [jsonschemastore](https://github.com/SchemaStore/schemastore/pull/4535) to add this

arhea:
@xiantang thats my understanding

drewdomi:
Please someone accept this one, I can't see this.

image

joekrill:
Are those required fields _actually_ required? I was under the impression _everything_ was optional. At least I don't have any problems running `air` with a completely empty config file. 

@drewdomi: As a temporary workaround you can add this schema directive to the top of your `.air.toml` file:

```toml
#:schema https://raw.githubusercontent.com/air-verse/air/0a3f56961f682c08c2d012dbf828188b4e12e854/air.schema.json
```

or basically disable schema checking by setting it to "any":

```toml
#:schema https://json.schemastore.org/any.json
```

MrBns:
Why this one is not getting approved ?

air.schema.json
```diff
@@ -0,0 +1,208 @@
...
+          "description": "Log file path, relative to tmp_dir."
+        },
+        "poll": {
+          "type": "boolean",
+          "description": "Use polling to detect file changes instead of fsnotify."
+        },
+        "poll_interval": {
+          "type": "integer",
+          "description": "Interval in milliseconds for polling file changes.",
+          "minimum": 500
```
MrBns: poll_interval can be 0. when I first installed air and initialized, I saw default value is 0.

xiantang:
close, cause this issue https://github.com/air-verse/air/issues/802

spotdemo4:
> close, cause this issue #802

Not sure why this was closed? That PR fixes the problems that arise if no schema is provided, but surely providing a schema is the better solution?

xiantang:
if we add a schema file, every time we add more options we also need to
edit the schema file, it will increase your maintenance load

PR #745, opened 2025-02-24
Feat: Use host header for proxy

Chrisischris:
My app serves multiple subdomains so for local development I access it with sub1.localhost:port for example. Since the proxy has localhost hardcoded you couldn't use live reload with these urls, this new config allows you to use the host header when proxying the request. I've also added a config for number of retries before failing since apps can take longer than a second to build.


slunak:
I have exactly the same use case. The host should not be hardcoded.

PR #747, opened 2025-02-28
fix: 修复命令行参数-build.full_bin无效的问题

sdfwds4:


xiantang:
please fix lint first

PR #748, opened 2025-03-03
fix: 修复命令行参数-build.full_bin无效的问题

sdfwds4:


runner/config.go
```diff
@@ -394,4 +394,12 @@ func (c *Config) WithArgs(args map[string]TomlInfo) {
 			setValue2Struct(v, value.fieldPath, *value.Value)
 		}
 	}
+
+	if len(c.Build.FullBin) > 0 {
```
xiantang, line 398: logic duplicated https://github.com/air-verse/air/blob/ad8fbb7a1cfba030dc09ae592b2fdc1c2eac6d15/runner/config.go#L324-L331

xiantang:
can we just extract https://github.com/air-verse/air/blob/ad8fbb7a1cfba030dc09ae592b2fdc1c2eac6d15/runner/config.go#L324-L331 to a function add put after https://github.com/air-verse/air/blob/ad8fbb7a1cfba030dc09ae592b2fdc1c2eac6d15/main.go#L97

sdfwds4:
```
func (c *Config) PostProcess() error {
	if len(c.Build.FullBin) > 0 {
		c.Build.Bin = c.Build.FullBin
		return nil
	}
	// Fix windows CMD processor
	var err error
	return err
}
```
```
cfg.WithArgs(cmdArgs) 
cfg.PostProcess()          // 给cfg增加额外的对外接口，增加使用复杂度
```

假如我像上面那样修改，PostProcess的功能仅是对c.Build.FullBin和filepath.Abs的简单封装，会给cfg增加额外的对外接口，增加使用复杂度，就不符合KISS原则了

xiantang:
> ```
> func (c *Config) PostProcess() error {
> 	if len(c.Build.FullBin) > 0 {
> 		c.Build.Bin = c.Build.FullBin
> 		return nil
> 	}
> 	// Fix windows CMD processor
> 	// CMD will not recognize relative path like ./tmp/server
> 	var err error
> 	c.Build.Bin, err = filepath.Abs(c.Build.Bin)
> 	return err
> }
> ```
> 
> ```
> cfg.WithArgs(cmdArgs) 
> cfg.PostProcess()          // 给cfg增加额外的对外接口，增加使用复杂度
> ```
> 
> 假如我像上面那样修改，PostProcess的功能仅是对c.Build.FullBin和filepath.Abs的简单封装，会给cfg增加额外的对外接口，增加使用复杂度，就不符合KISS原则了

sorry i need to write in English, cause there are contributors all around the world. and i made WithArgs private, and hide logic in preprocess function. i think this way should be better.

Issue #730, opened 2025-01-20
-build.exclude_regex argument isn't working


condemil:
Here is the command to be able to reproduce the issue:

```sh
air -build.include_ext "go,templ" -build.exclude_regex "_templ.go" -build.cmd "templ generate && go build -o ./tmp/main ."
```

When the command is running there is infinite loop of re-runs:

```sh
$ air -build.include_ext "go,templ" -build.exclude_regex "_templ.go" -build.cmd "templ generate && go build -o ./tmp/main ."

  __    _   ___  
 / /\  | | | |_) 
/_/--\ |_| |_| \_ v1.61.7, built with Go go1.23.4

watching .
watching static
watching templates
!exclude tmp
building...
(✓) Complete [ updates=4 duration=1.063292ms ]
running...
2025/01/20 20:42:28 Starting server on :8080
templates/base_templ.go has changed
building...
(✓) Complete [ updates=4 duration=1.217167ms ]
running...
2025/01/20 20:42:29 Starting server on :8080
templates/base_templ.go has changed
...
```

When I use `.air.toml` file with `exclude_regex = ["_templ.go"]` I don't have such issue, so I assume there is some issue with command line argument.

xiantang:
https://github.com/air-verse/air/pull/748 solved

patrickkdev:
When will this be released?

PR #750, opened 2025-03-05
Don't ignore build.exclude_regex supplied via command line arguments

victor-tsykanov:
Currently air ignores `build.exclude_regex` supplied as command line argument. The problem is mentioned in https://github.com/air-verse/air/issues/730.

### What happens
Internally air uses precompiled exclusion regex. Regex is compiled when Config is initialized. Command line arguments are applied after initialization, and as a result precompiled exclusion regex is not updated.

### Solution
Recompile exclusion regex after command line arguments are applied to the config.

### DIFF
--- a/runner/config.go
+++ b/runner/config.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"log"
 	"os"
 	"path/filepath"
 	"reflect"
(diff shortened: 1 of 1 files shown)


xiantang:
let me find a good way to fix this problem, i dont really wanna run code two times

xiantang:
should be fix by https://github.com/air-verse/air/pull/748

Issue #754, opened 2025-03-20
Proxy blocking when reaching browser SSE stream limit


grelland:
Hi, and thanks for a great tool!

I'm using the proxy whilst developing a basic template-based web application. It works mostly well, but I noticed that on every seventh page load the browser will hang for about a minute before the request is handled.

After some debugging I traced the problem to be the per-browser/hostname SSE stream limit (which just so happens to be exactly **6**), coupled with the cleanup logic in the `reloadHandler` [here](https://github.com/air-verse/air/blob/326e87d6fbe92007dd65affd89616f19806eb0f4/runner/proxy.go#L178-L181). This blocks until the (dropped) connection is eventually cleaned up and the browser is able to establish the `EventSource`.

I have a fix in mind and will work on a PR  👍

michaeldebetaz:
@grelland Thank you so much for opening this issue. It was really annoying and I was totally lost on why only the proxy would hang after navigating a couple of times.

I am playing with Datastart (which uses server-side event) and now I understand it comes from the limited number of connections.

Your fix is good. In the meantime, I will try to get the proxy to use HTTP 2 (which is has no limitations on event/stream) as far as I know.

EDIT : trying to use HTTP/2 with air proxy is a whole mess. I think I'll stick to triggering event source cleanup from now on.

PR #755, opened 2025-03-20
754 - Clean up EventSource when the page is closed/reloaded

grelland:
Fixes [issue 754](https://github.com/air-verse/air/issues/754)

This is a very basic fix but it works. 

I did try to implement a more robust handling of the SSE connection on the server side as well (SSE heartbeats, lowering TCP timeouts), but could not get it to work reliably without also risking making the proxy far less transparent.

### DIFF
--- a/runner/proxy.js
+++ b/runner/proxy.js
@@ -1,6 +1,10 @@
 (() => {
     const eventSource = new EventSource("/__air_internal/sse");
 
+    window.addEventListener('beforeunload', function() {
+        eventSource.close();
+    })
+
     eventSource.addEventListener('reload', () => {
         location.reload();
     });


ofipify:
I keep running into this as well. We have a page that executes an XHR call to a subsequent resource and I've noticed the occasional hang as the browser runs into its SSE connection limits.

ofipify (approved):

grelland:
Any chance this gets merged?

xiantang:
seems not fix the issue

xiantang:
https://github.com/air-verse/air-reproducible-example/tree/main/issue-754-proxy-sse-limit @grelland, this is the example

grelland:
> seems not fix the issue
> https://github.com/air-verse/air-reproducible-example/tree/main/issue-754-proxy-sse-limit

The reproduction scenario says to open 7 tabs. If they're all kept open you are correct, this fix doesn't apply. It only runs `beforeunload`, i.e. when the tab is closed. It _does_ solve the problem where an already closed tab is hogging a slot, though, which I suspect is what most real world users will hit...
To support 6+N tabs open I suspect a more sophisticated fix is needed. Probably could use a `SharedWorker` so the tabs share an `EventSource`. I'll try to find some time to implement this.

ofipify:
I think @grelland's fix is a pragmatic workaround for the issue reported in #754. A better test would be to open and close 7 windows in short succession via `window.open`.

xiantang:
LGTM

ofipify:
Thank you @xiantang for merging and @grelland for the fix. :pray:

grelland:
@xiantang Thanks for merging!

In the meantime I got curious and went ahead with the SharedWorker based implementation. It solves both problems now, but adds a fair bit of code 🤷‍♂️  https://github.com/air-verse/air/pull/859

xiantang:
I don’t mind to add bit of code, but please do not break the behaviour

PR #762, opened 2025-04-16
chore: update golangci-lint to 1.64

MoritzKrafeld:
Once we are on version 1.64 for golangci-lint, we should also be able to update to the latest go version (1.24). Once that is also done, we can migrate golangci-lint from version 1.x to 2.x :smile:


MoritzKrafeld:
Hey @linzhengen, would you be so kind and take a look at this MR? Once that is done, we can go ahead and update to go 1.24 :smile:

Issue #737, opened 2025-02-05
docs: Air **requires** tty, a detail that may be more important for Docker users


aentwist:
## Description

Air runs my program, but does not live reload, when using Docker. It works on the host machine.

As a... for now... WSL developer, I **really** tried to get it working however I could.

Here's what I've tried:

##### Host machines

- Windows 11 / WSL2 Debian 12 (amd64) [Docker Desktop]
- AWS EC2 Ubuntu 24.04.1 (amd64) [Docker engine]

-> does not appear to be a host issue (WSL/Windows/Docker Desktop)

##### Docker configurations

- copy code into container. `docker run`
- copy code into container. `docker compose up`
- bind mount code into container. `docker run`
- bind mount code into container. `docker compose up`

-> does not appear to be either a bind mount or docker compose issue

##### Go versions

- 1.23.6 (`go install air` + `air -c .air.toml`)
- 1.24rc2 (`go get -tool air` + `go tool air -c .air.toml`)

-> does not appear to be incompatibility with the upcoming `go get -tool` + `go tool` or other unstable release bug

##### Air versions

- 1.61.7
- 1.60.0
- 1.51.0

-> does not appear to be a regression in Air

##### Enabling `poll = true`

-> I mean it doesn't seem like a fsnotify bug... because of all of the above, and because this isn't working for me either...

## Reproduction

https://github.com/aentwist/air-repro

## Environment / Preferred Environment

- linux + docker engine - sanity
- Docker version 27.5.1, build 9f9e405 (latest installed via Docker's apt repo)
- go 1.24rc2 - really excited to use the new dev dependencies feature
- air 1.61.7

aentwist:
Related to #190 #26 
Potentially related to #272

aentwist:
##### First problem

- go embeds vcs information into binaries
- when copying files, ownership of .git is transferred to the docker user
- when bind mounting volumes, ownership is not transferred
- git has a safety mechanism [`safe.directory`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-safedirectory)
  - prevents interacting with non-owned git data (because all git data should always be owned)
  - causes go build to fail - cannot embed vcs information (`error obtaining VCS status: exit status 128 Use -buildvcs=false to disable VCS stamping.`)
  - try `git status` to check whether this is being triggered

##### Solution

- can tell go to not embed vcs information. this is ok to do for development only
  - add `-buildvcs=false` to build command
[...]

PR #768, opened 2025-05-09
feat: add force color backed by config flag

kema-dev:
## Why

- Be able to force colorful output

## How

- Add a `force` flag to color config and set `color.NoColor` accordingly

## Related issues

- #737

### DIFF
--- a/runner/config.go
+++ b/runner/config.go
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	"dario.cat/mergo"
+	"github.com/fatih/color"
 	"github.com/pelletier/go-toml"
 )
 
@@ -163,6 +164,7 @@ type cfgColor struct {
 	Watcher string `toml:"watcher" usage:"Customize watcher part's color"`
 	Build   string `toml:"build" usage:"Customize build part's color"`
 	Runner  string `toml:"runner" usage:"Customize runner part's color"`
+	Force   bool   `toml:"force" usage:"Force colorized output"`
 	App     string `toml:"app"`
 }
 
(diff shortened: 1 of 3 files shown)


runner/config.go
```diff
@@ -334,6 +336,11 @@ func (c *Config) preprocess(args map[string]TomlInfo) error {
 	// CMD will not recognize relative path like ./tmp/server
 	c.Build.Bin, err = filepath.Abs(c.Build.Bin)
 
+	if c.Color.Force == true {
+		// Force colorful output, see https://github.com/fatih/color#disableenable-color
+		color.NoColor = false
+	}
```
ccoVeille, line 342: Usually this kind of configuration flag is done via a string that accepts the following values: auto (default), always, never


```suggestion
	switch c.Color.Mode {
		case "always":
			color.NoColor = false
		case "never":
			color.NoColor = true
		case "auto", "":
			// expected default value
		default:
			return nil, fmt.Errorf("unsupported value for color mode: %s. Expected always, auto, or never", c.Color.Mode)
	}
```

This way people can temporarily disable the color via the configuration.

I know fatih lib supports NO_COLOR, but people expect configuration to control the app without using env

kema-dev: LGTM :+1: [`cb701b7` (#768)](https://github.com/air-verse/air/pull/768/commits/cb701b7b319f1ac628a9d97e3954562b8c0bdfa5)

ccoVeille (approved):

xiantang:
@codex review

xiantang:
@codex review

xiantang:
close, already include in master branch

PR #779, opened 2025-07-12
docs: add `go get -tool` install method in README

alexandregv:
Since Go 1.24, go tools can be added to the `go.mod` with the new `tool` directive and `go get -tool` command.  
This allows to download the tool automatically like normal dependencies, and to specify a version specific to each project.

- Go 1.24 releases notes about this: https://go.dev/doc/go1.24#go-command
-  `go get -tool` and `go tool` documentation: https://go.dev/doc/modules/managing-dependencies#tools

In my opinion this would be the new recommended way, but without other opinions I kept the `go install` as the recommended one.

### DIFF
--- a/README-ja.md
+++ b/README-ja.md
@@ -62,6 +62,16 @@ go 1.23以上を使う場合:
 go install github.com/air-verse/air@latest
 ```
 
+### `go get -tool` を使う場合
+
+go 1.24以上を使う場合:
+
+```bash
+go get -tool github.com/air-verse/air@latest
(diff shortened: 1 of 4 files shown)


ofipify:
Awesome! This would be a great addition to the documentation, avoiding the hassle of having to install and invoke from the Go binary directory, which might not be in your path (yet) or if you use something like `asdf` for managing multiple versions.

xiantang:
lets go thanks

PR #797, opened 2025-09-07
chore: update Go version to 1.25.1 and dependencies

cosmtrek:
- Updated Dockerfile to use Go 1.25.1 for both builder and final images.
- Modified README files (Japanese, Chinese Simplified, Chinese Traditional, and English) to reflect the new Go version requirement of 1.25.1.
- Updated go.mod to specify Go 1.25 and updated dependencies:
  - Upgraded hugo from v0.147.6 to v0.149.1
  - Upgraded testify from v1.10.0 to v1.11.1
  - Upgraded other indirect dependencies to their latest versions.
- Updated go.sum to reflect the changes in dependencies.


reneleonhardt:
What does the go directive have to do with the go release used for building binaries or Docker images?
When I analyze all dependencies the maximum go directive is 1.24.0 to satisfy compatibility.
Interestingly enough, that's the exact version `go mod tidy` produces when you use the previous go directive (or an even older release like `go 1.23` for double-checking).

```shell
$ go version
go version go1.24.7
$ go mod tidy
```
Changes the previous `go 1.24` to `go 1.24.0` to stay compatible to all dependencies.

All new versions in Dockerfile and workflows already ensure that go 1.25.1 is used to build binaries, as you can verify easily:
```shell
$ strings air_1.63.0_darwin_amd64 | grep "^go1\.2" | head
go1.25.1
```

reneleonhardt:
Keep in mind that golangci-lint could break your CI unnecessarily, maybe for some workflows or events it shouldn't be allowed to break.
https://www.linkedin.com/pulse/15-hours-one-linter-error-hidden-cost-mismatched-ci-viraj-ghevariya-hb4mc

Issue #796, opened 2025-09-02
air 启动二进制文件总是很慢（Air is always slow to start binaries.）


liushuai05:
不得不说air 确实是一个非常棒的项目，但是我在使用中遇到一个问题，就是每次air启动二进制文件都跟蜗牛一样，在running...后都要等到3-5秒钟（甚至慢的时候要8秒）,可能很多兄弟不会在意这个时间，但是我们公司是做快开发的，这一点点时间可能对于我们都很重要，所以我研究了一下大佬你的代码,"github.com/creack/pty" 这个包引起了我的注意，在AI的解释下，我得以知道它是采用的虚拟终端"github.com/creack/pty" 这个包引起了我的注意，在AI的解释下，我得以知道它是采用的虚拟终端,它会带来额外开销，我改了一下源码编译并替换掉我电脑上的air，是的它运行编译后的二进制变得飞快，我相信你使用pty肯定有你的原因，但是我希望你能给用户一个选择，比如在配置中维护一个配置，如果这个配置给true那么使用pty（当然你可以让他默认为true），如果这个配置给false那么使用golang原生的stdout和stderr
(Air is really a great project, but I have a problem when I use it, that is, every time air starts the binary file, it will take 3-5 seconds after running ... Maybe many brothers won't care about this time, but our company is developing fast, and this little time may be very important to us, so I studied your code,The package "github.com/creack/pty" caught my attention, and under the explanation of AI, I was able to know that it was a virtual terminal，The package "github.com/creack/pty" caught my attention. Under the explanation of AI, I was able to know that it was a virtual terminal, which would bring extra expenses. I changed the source code compilation and replaced the air on my computer. Yes, it ran the compiled binary quickly，I believe you must have your reasons for using pty, but I hope you can give users a choice, such as maintaining a configuration in the configuration, and using pty if this configuration is true (of course, you can let him default to true)，If this configuration is set to false, use golang's native stdout and stderr.)

下面是我修改的代码（Here is the code I modified.）:
```
func (e *Engine) startCmd(cmd string) (*exec.Cmd, io.ReadCloser, io.ReadCloser, error) {
	c := exec.Command("/bin/sh", "-c", cmd)
	// 创建管道，将子进程的stdout和stderr连接到父进程
	stdoutPipe, err := c.StdoutPipe()
	if err != nil {
		return nil, nil, nil, err
	}
	stderrPipe, err := c.StderrPipe()
	if err != nil {
		return nil, nil, nil, err
	}
[...]

PR #544, opened 2024-03-05
add no_pty conifg option to fix #533

graf0:
should fix ticket #533

Issue #533, opened 2024-02-21
Unescaped color output when running?


hahuang65:
Getting this output

`^[]10;rgb:cdcd/d6d6/f4f4^[\^[[32;1R^[]11;rgb:1e1e/1e1e/2e2e^[\^[[32;36R` when running.

Tried on multiple terminal emulators.

```
go run github.com/cosmtrek/air@latest --build.cmd "go build -o /tmp/bin/foobar ./cmd/api" --build.bin "/tmp/bin/foobar"

  __    _   ___  
 / /\  | | | |_) 
/_/--\ |_| |_| \_ v1.50.0, built with Go go1.22.0

watching .
watching cmd
watching cmd/api
watching docker
watching docker/docker
watching docker/docker/prometheus.yml
watching internal
watching internal/env
watching internal/log
watching internal/otel
watching internal/request
watching internal/response
watching internal/validator
watching internal/version
watching prometheus.yml
!exclude tmp
building...
running...
^[]10;rgb:cdcd/d6d6/f4f4^[\^[[32;1R^[]11;rgb:1e1e/1e1e/2e2e^[\^[[32;36R2024-02-21 08:58:28 INFO <otel/otel.go:26> setting up telemetry endpoint=localhost:4317
2024-02-21 08:58:28 INFO <api/server.go:61> starting server server="[addr=:4444]"
```

MKAbdElrahman:
I have the same issue.

enrichman:
Yep, same. Everything is fine on my Mac, but on Linux I have this problem.

graf0:
same here

graf0:
It seems that problem is with using github.com/creack/pty in util_linux.go. And it's not caused by ie.: logging via log or slog. But after adding github.com/charmbracelet/wish I've got that strange effect. 
I failed to diagnose which package is problematic. But in general - it seems that running command via pty is main problem. 

Because by definition most program run by air are not pty programs ,but rather services/servers - I suggest adding no_pty optoin in such cases. So if you have problems with strage output like this:

```
^[]10;rgb:cdcd/d6d6/f4f4^[\^[[32;1R^[]11;rgb:1e1e/1e1e/2e2e^[\^[[32;36R
```

Set such option in .air.toml:

```toml
[screen]
  no_pty = true
```

That fixes problem.

hahuang65:
For what it's worth, another binary recently fixed this:

https://github.com/evilmartians/lefthook/releases/tag/v1.6.3

[...]

PR #809, opened 2025-11-09
Stop using pty on Linux

xiantang:
This solution for Linux will identify all child processes and attempt to terminate them. For the tty package, the old implement, it will start a new pseudo terminal, allowing all child processes to be organized under that tree. In this implementation, we use a recursive approach to locate all child processes and then terminate them one by one. This method will also eliminate processes that have different process group IDs (PGIDs).
close https://github.com/air-verse/air/issues/796
https://github.com/air-verse/air/pull/544
https://github.com/air-verse/air/issues/533


xiantang:
we can't remove pty dirctly, cause once i remved, test failed, air can't kill all subprocess correctly

PR #817, opened 2025-11-15
feat: add option to skip deleting the binary on rebuild

steveiliop56:
Hello,

I noticed a small issue with the way air works. In my setup I am using delve to run the binary so as I can attach from my editor and debug. The issue is that delve gets interpreted as the build out and gets deleted on rebuild resulting in a `file not found` error when it tries to run it again. My current solution is to use `cp /go/bin/dlv dlv` as a pre-run cmd but I believe the solution this PR proposes is much better. I am adding a `delete_bin` option that can be set to `false` so as the binary will not be deleted allowing the user of debugger programs or wrappers to execute the main program. This will also cause any issues with go since it overwrites the binary by default. Let me know if you need anything changed.

### DIFF
--- a/runner/config.go
+++ b/runner/config.go
@@ -60,6 +60,7 @@ type cfgBuild struct {
 	KillDelay        time.Duration `toml:"kill_delay" usage:"Delay after sending Interrupt signal"`
 	Rerun            bool          `toml:"rerun" usage:"Rerun binary or not"`
 	RerunDelay       int           `toml:"rerun_delay" usage:"Delay after each execution"`
+	DeleteBin        bool          `toml:"delete_bin" usage:"Delete binary before building"`
 	regexCompiled    []*regexp.Regexp
 }
 
@@ -225,6 +226,7 @@ func defaultConfig() Config {
 		Delay:        1000,
 		Rerun:        false,
 		RerunDelay:   500,
+		DeleteBin:    true,
 	}
 	if runtime.GOOS == PlatformWindows {
 		build.Bin = `tmp\main.exe`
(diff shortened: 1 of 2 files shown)


xiantang:
I didn't really understand this feature. Why don't you just put the dlv in PATH?

steveiliop56:
I am using `full_bin` and wherever I put the delve binary air will delete it on rebuild.

xiantang:
image
for my case it look good,
image
and my dlv is under /home/neo/go/bin/dlv

my config is 
```
root = "."
testdata_dir = "testdata"
tmp_dir = "tmp"

[build]
  args_bin = []
  bin = "./tmp/main"
  # cmd = "go build -o ./tmp/main ."
  delay = 1000
  exclude_dir = ["assets", "tmp", "vendor", "testdata"]
  exclude_file = []
  exclude_regex = ["_test.go"]
  exclude_unchanged = false
  follow_symlink = false
  full_bin = "dlv exec --accept-multiclient --log --headless --continue --listen=:8999 --api-version 2 ./tmp/main"
  include_dir = []
  include_file = []
  kill_delay = "0s"
  log = "build-errors.log"
  poll = false
  poll_interval = 0
  post_cmd = []
  pre_cmd = []
  rerun = false
  rerun_delay = 500
  send_interrupt = false
  stop_on_error = false

[color]
  app = ""
  build = "yellow"
  main = "magenta"
  runner = "green"
  watcher = "cyan"

[log]
  main_only = false
  # silent = true
  time = false

[misc]
  clean_on_exit = false

[proxy]
  app_port = 0
  enabled = false
  proxy_port = 0

[screen]
  clear_on_rebuild = false
  keep_scroll = true
```

steveiliop56:
Oh, you are correct. Turns out I had set `bin` to delve instead of the build out :facepalm:.

Issue #819, opened 2025-11-15
Watching for changes in parent directories


alexflint:
It seems that if I write `air --build.include_dir="../pkg"`, air does not in fact watch the directory at `../pkg`, because (as far as I can tell -- correct me if I'm wrong) only directories within the current directory are considered for watching. Any chance this could be amended?

xiantang:
I'm looking into this issue, thanks for reporting.

xiantang:
https://github.com/air-verse/air/pull/820
I have a PR now, could you try this PR by `go install github.com/air-verse/air@fix_exclude_dir` ?

alexflint:
Hey, looks great! I tried that PR and it works for me. Will comment also on the PR. Thank you!

PR #820, opened 2025-11-16
Handle include_dir outside root

xiantang:
close https://github.com/air-verse/air/issues/819


alexflint:
Tried this PR locally and confirmed that it works for the use case that I had in mind when reporting #819. Thank you!

PR #840, opened 2025-12-30
fix: Race condition causes builds to cancel themselves when triggered during active build

### LINKED ISSUE #784: Race condition causes builds to cancel themselves when triggered during active build
# Description

When a file change triggers a build while an existing build is running, a race condition in `engine.go` causes the new build to cancel itself, making the running binary outdated.

# Root cause

You make a change to trigger build A. In the `buildRun()` function, build A will put `true` onto `buildRunCh`, and then proceed with its build. While it is building, you make a change to trigger build B. In the `start()` function, build B sees that there is a value on `buildRunCh`, so it puts a value onto `buildRunStopCh`:

```go
// already build and run now
select {
case <-e.buildRunCh:
  e.buildRunStopCh <- true
default:
}
```

In then calls `go buildRun()`. Since `buildRunCh` is now empty, it will be able to place `true` onto it. Directly after, it will see that `buildRunStopCh` is not-empty, and will **effectively cancel itself.**

```go
select {
case <-e.buildRunStopCh:
  return
default:
}
```

# Reproduction

Place the following in your `.air.toml` file to make the build process take a while. Make a change, wait a little bit (less than 10 seconds), and then make another change - the second one will not be reflected in the running code.

```toml
cmd = "echo 'Build started at' $(date +%H:%M:%S) && go build -o ./tmp/main . && echo 'Build complete' && sleep 10"
```

majiayu000:
## Summary
This PR fixes #784

## Changes
- Added `buildGeneration` atomic counter to assign each build a unique generation number
- Changed `buildRunStopCh` from `chan bool` to `chan uint64` to carry the target generation with stop signals
- Added `buildRunStopCheckMu` mutex to ensure atomic operations when checking/sending stop signals
- When a stop signal is sent, it now includes the current generation number
- When a build receives a stop signal, it only stops if the signal's generation >= its own generation (meaning the signal was meant for it or a newer build that hasn't started yet)
- Added regression test `TestBuildRunRaceCondition` to verify the fix

## Root Cause
The race condition occurred when:
1. Build A was running
2. A file change triggered a new build
3. A stop signal was sent to channel for Build A
4. Build B started before Build A consumed the stop signal
5. Build B consumed the stop signal meant for Build A and cancelled itself

## Solution
By associating each stop signal with a target generation number, new builds can distinguish between stop signals meant for them vs. older builds. A build only stops if the stop signal's generation >= its own generation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/runner/engine.go
+++ b/runner/engine.go
@@ -29,13 +29,14 @@ type Engine struct {
 
 	eventCh       chan string
 	watcherStopCh chan bool
-	buildRunCh    chan bool
-	// buildGeneration is used to prevent race conditions where a new build
-	// receives a stop signal meant for a previous build. Each build is assigned
-	// a generation number, and stop signals include the target generation.
-	buildGeneration     atomic.Uint64
-	buildRunStopCh      chan uint64
-	buildRunStopCheckMu sync.Mutex
+	// buildRunCh serves dual purpose:
+	// 1. As a semaphore ensuring only one build runs at a time (buffer size 1)
+	// 2. Carries each build's unique stop channel for cancellation
+	// When a new build starts, it retrieves the previous build's stop channel,
+	// closes it to signal cancellation, then inserts its own fresh channel.
+	// This prevents the race condition where a new build could consume a stop
+	// signal meant for a previous build (issue #784).
+	buildRunCh chan chan struct{}
 	// binStopCh is a channel for process termination control
 	// Type chan<- chan int indicates it's a send-only channel that transmits another channel(chan int)
 	binStopCh chan<- chan int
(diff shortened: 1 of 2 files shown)


xiantang:
i don't think this is a good fix, it introduces the mutex and a new field

majiayu000:
我补了一个更稳的方案，核心是单线程 build worker + 事件合并 + 可取消 build（context）。\n\n改动要点：\n- 触发构建只发请求，buildLoop 串行执行，避免并发 build 竞态。\n- 新事件到来会取消当前 build（context cancel），pre_cmd/build 都会在取消后提前退出。\n- build 完成但已有新请求时不 run bin，保证只运行最新产物。\n\n实现细节：\n- 新增 buildReqCh/buildPending + buildLoop/requestBuild。\n- building 使用 context 可取消，新增 startCmdWithContext 与 runCommandCopyOutputContext。\n- 补充两个测试：buildPending 跳过 build、以及 context cancel 的命令终止（Windows 下跳过）。\n\n如果你希望保持原有架构，也可只采纳“单线程调度 + buildPending”的思路，取消逻辑可延后再细化。

xiantang:
> 我补了一个更稳的方案，核心是单线程 build worker + 事件合并 + 可取消 build（context）。\n\n改动要点：\n- 触发构建只发请求，buildLoop 串行执行，避免并发 build 竞态。\n- 新事件到来会取消当前 build（context cancel），pre_cmd/build 都会在取消后提前退出。\n- build 完成但已有新请求时不 run bin，保证只运行最新产物。\n\n实现细节：\n- 新增 buildReqCh/buildPending + buildLoop/requestBuild。\n- building 使用 context 可取消，新增 startCmdWithContext 与 runCommandCopyOutputContext。\n- 补充两个测试：buildPending 跳过 build、以及 context cancel 的命令终止（Windows 下跳过）。\n\n如果你希望保持原有架构，也可只采纳“单线程调度 + buildPending”的思路，取消逻辑可延后再细化。

https://github.com/air-verse/air/pull/840/commits/5576e01c52ef30f892f981eba7521da5b12760ba fix already, u can check out this commit

PR #841, opened 2026-01-01
fix: Proxy should not repackage chunks from the app for HTTP 1.1 Transfer-Encoding: chunked

### LINKED ISSUE #791: Proxy should not repackage chunks from the app for HTTP 1.1 Transfer-Encoding: chunked
In case we have long-lived HTTP 1.1 SSE stream, the server will send events in chunks and flush them, but the proxy will accumulate and repackage them.

The Air's proxy accumulates data from server until it reaches first `0x200` bytes, and then after that accumulates data into chunks of `0x800` bytes.

This will cause untimely deliveries to client.

The proxy should send chunks verbatim as received from server, and flush outgoing buffer after each chunk (at least for `Content-Type: text/event-stream`, but possibly for any kind of chunked delivery).

Below is an example with long-lived SSE endpoint pushing events every 3 seconds. In this case the server emits events in chunks of  `0x50` bytes and flushes after every chunk. 

```
> curl --raw -i http://localhost:3002/ssetst
HTTP/1.1 200 OK
Cache-Control: no-cache
Connection: keep-alive
Content-Type: text/event-stream
Date: Mon, 18 Aug 2025 08:15:04 GMT
Transfer-Encoding: chunked

50
event: datastar-patch-signals
data: signals {"ssetstsignal":{"iteration":"0"}}


50
event: datastar-patch-signals
data: signals {"ssetstsignal":{"iteration":"1"}}


50
event: datastar-patch-signals
data: signals {"ssetstsignal":{"iteration":"2"}}


50
event: datastar-patch-signals
data: signals {"ssetstsignal":{"iteration":"3"}}

... keeps pushing events every 3 sec
```

Here is how same response comes through proxy. It comes out as `0x200` bytes chunk followed by `0x800` bytes. Instead of delivering every 3 seconds we will wait for `0x800 / 0x50 * 3sec` :
```
HTTP/1.1 200 OK
Access-Control-Allow-Origin: *
Cache-Control: no-cache
Connection: keep-alive
Content-Type: text/event-stream
Date: Mon, 18 Aug 2025 08:15:09 GMT
Via: HTTP/1.1 localhost:3082
Transfer-Encoding: chunked

200
event: datastar-patch-signals
data: signals {"ssetstsignal":{"iteration":"0"}}

event: datastar-patch-signals
data: signals {"ssetstsignal":{"iteration":"1"}}

event: datastar-patch-signals
data: signals {"ssetstsignal":{"iteration":"2"}}

... More data here

event: datastar-patch-signals
da
800
ta: signals {"ssetstsignal":{"iteration":"6"}}

event: datastar-patch-signals
data: signals {"ssetstsignal":{"iteration":"7"}}

event: datastar-patch-signals
data: signals {"ssetstsignal":{"iteration":"8"}}

... More data here

event: datastar-patch-signals
data: signals {"ssetstsignal"
5c8 # The last chunk is less than 0x800 bytes long
:{"iteration":"31"}}

event: datastar-patch-signals
data: signals {"ssetstsignal":{"iteration":"32"}}

...
```

majiayu000:
Fixes #791

## Changes
- Detect streaming responses (Transfer-Encoding: chunked or text/event-stream)
- Forward streaming responses with immediate write+flush instead of io.Copy buffering
- Fix WriteHeader placement to be called after setting headers
- Add SSE proxy test

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -5,6 +5,7 @@ Guidelines for contributors and AI coding agents working in this repository.
 ## Goals
 
 - Keep changes minimal, focused, and idiomatic to Go.
+- Code and comments must be write in English
 - Prefer root‑cause fixes over band‑aids; don’t refactor unrelated code.
 - Maintain user‑facing behavior and CLI flags unless explicitly changing them.
 
--- a/runner/proxy.go
+++ b/runner/proxy.go
@@ -141,40 +141,40 @@ func (p *Proxy) proxyHandler(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Access-Control-Allow-Origin", "*")
 	w.Header().Add("Via", viaHeaderValue)
 
-	// Determine if this is a streaming response that should be forwarded verbatim
-	isStreaming := resp.Header.Get("Transfer-Encoding") == "chunked" ||
-		strings.Contains(resp.Header.Get("Content-Type"), "text/event-stream")
+	// Determine if this is a streaming response
+	streaming := isStreamingResponse(resp)
 
+	// Handle non-HTML responses
 	if !strings.Contains(resp.Header.Get("Content-Type"), "text/html") {
-		if !isStreaming {
-			w.Header().Set("Content-Length", resp.Header.Get("Content-Length"))
-		}
-		w.WriteHeader(resp.StatusCode)
-		if isStreaming {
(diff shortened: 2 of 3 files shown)


xiantang:
local test good, let me review the code

xiantang:
im going to create a new PR and include your commits

xiantang:
tks

PR #859, opened 2026-01-12
754  use shared EventSource

grelland:
More sophisticated approach to the problem attempted solved in https://github.com/air-verse/air/pull/755
close https://github.com/air-verse/air/pull/754

Uses a SharedWorker to multiplex the SSE stream rather than maintaining an EventSource per tab/window.
The [reproducible example](https://github.com/air-verse/air-reproducible-example/tree/main/issue-754-proxy-sse-limit) now passes properly.

### DIFF
--- a/runner/proxy.js
+++ b/runner/proxy.js
@@ -1,6 +1,12 @@
 (() => {
     let worker = null;
 
+    const disconnectWorker = () => {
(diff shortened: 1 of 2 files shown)


xiantang:
great, let me review it this week

PR #862, opened 2026-01-14
feat: Add option to ignore dangerous root dir protections

### LINKED ISSUE #861: Add option to explicitly override "dangerous root directories" protections
**Is your feature request related to a problem? Please describe.**
After the release of [v1.63.8](https://github.com/air-verse/air/releases/tag/v1.63.8), my development environment - which uses `air` to provide live reload - does not build or run my project. My project is built and run within a Docker container and the user running the binary has its home set to the project's location on disk. As such, there are few files in the user's home directory. For example, my app user has its home set to `/app` and the entire project directory is copied into that location for building and running within the container. 

**Describe the solution you'd like**
An explicit option to ignore or override the protections for dangerous root directories. This option can come with an explicit warning that using it is not recommended.

**Describe alternatives you've considered**
* Pinning my tooling to [v1.63.7](https://github.com/air-verse/air/releases/tag/v1.63.7), which is undesirable as it would prevent receiving updates to the tool.
* Updating my development and build process to change the location of the app user's home and the project's directory. This is undesirable as this would require updating other parts of my project and tooling to align with this new behavior of `air`.

**Additional context**
Related to previous PR #851 and issue #761.
I am willing to contribute to implement this functionality.

abborg:
This PR adds a new configuration option in the `build` section called `ignore_dangerous_root_dir`. This option when set to `true` overrides the dangerous root protections and allows air to build and execute the entrypoint even if the `root` dir is set to `/`, `~`, or `/root`. This option is not recommended without careful consideration due to the potential to have performance impact from file watchers.

Resolves #861

### DIFF
--- a/air_example.toml
+++ b/air_example.toml
@@ -36,8 +36,8 @@ exclude_file = []
 exclude_regex = ["_test\\.go"]
 # Exclude unchanged files.
 exclude_unchanged = true
-# Override dangerous root directory protections. (Not recommended)
-ignore_dangerous_root_dir = true
+# Ignore dangerous root directory that could cause excessive file watching
+ignore_dangerous_root_dir = false
 # Follow symlink for directories
 follow_symlink = true
 # This log file is placed in your tmp_dir.
--- a/runner/config.go
+++ b/runner/config.go
@@ -381,7 +381,7 @@ func (c *Config) preprocess(args map[string]TomlInfo) error {
 		if !c.Build.IgnoreDangerousRootDir {
 			return fmt.Errorf("refusing to run in %s - this would watch too many files. Please run air in a project directory", dirName)
 		}
-		fmt.Fprintln(os.Stdout, "[warning] ignoring root directory protections. This could cause excessive file watching. It is recommended to run air in a project directory", dirName)
+		fmt.Fprintln(os.Stdout, "[warning] ignoring root directory protections. This could cause excessive file watching. It is recommended to run air in a project directory")
 	}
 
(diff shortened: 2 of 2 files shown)


runner/config.go
```diff
@@ -377,7 +378,10 @@ func (c *Config) preprocess(args map[string]TomlInfo) error {
 
 	// Check for dangerous root directories that could cause excessive file watching
 	if isDangerous, dirName := isDangerousRoot(c.Root); isDangerous {
-		return fmt.Errorf("refusing to run in %s - this would watch too many files. Please run air in a project directory", dirName)
+		if !c.Build.IgnoreDangerousRootDir {
+			return fmt.Errorf("refusing to run in %s - this would watch too many files. Please run air in a project directory", dirName)
+		}
+		fmt.Fprintln(os.Stdout, "[warning] ignoring root directory protections. This could cause excessive file watching. It is recommended to run air in a project directory", dirName)
```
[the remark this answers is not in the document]
xiantang: ah, ok its fine.

air_example.toml
```diff
@@ -36,6 +36,8 @@ exclude_file = []
 exclude_regex = ["_test\\.go"]
 # Exclude unchanged files.
 exclude_unchanged = true
+# Override dangerous root directory protections. (Not recommended)
+ignore_dangerous_root_dir = true
```
xiantang: if default its false, so we need to set there as false

xiantang: so there should set as false

xiantang:
tks

Issue #279, opened 2022-05-21
cross-platform config file


batara666:
default air.toml point to `./tmp/main` while It's need to be `.\\tmp\\main.exe` on windows

chjiyun:
CMD `go build -o ./tmp/main .`  is unavailable on windows

Nikola-Milovic:
Any progress on this?

snowdream:
Any News？

xiantang:
will continue work on this

FrancescoLuzzi:
Using `powershell` you can run an application with no extensions by launching it in a process:
```
Start-Process -FilePath "./path/to/app" -ArgumentList "-arg test" -Wait -NoNewWindow
```

More infos (and a more powershell black belt approach) in this [stack overflow response](https://stackoverflow.com/questions/77075489/powershell-run-executable-without-extension#answer-77077461)

xiantang:
pls try the new cross-platform config
eg:

```
# Platform-specific build overrides
[build.windows]
cmd = "go build -o ./tmp/main.exe ."
bin = "tmp\\main.exe"
entrypoint = ["tmp\\main.exe"]
```

Issue #767, opened 2025-05-04
`include_dir` in air.toml expects OS specifc path notation?


dw-0:
I just came across a weird config behavior.
My project is an nx monorepo, combining frontend and backend code. I make use of the common approach of splitting code into apps and libs. That's where im required to specify which directories air should watch, because my project structure looks like this:
```
project-root/
├── apps/
│   ├── backend/
│   │   ├── .air.toml
│   │   ├── cmd/
│   │   ├── internal/
│   │   ├── project.json
│   │   ├── README.md
│   │   └── tmp/
│   └── frontend/
├── libs/
│   ├── backend/
│   └── frontend/
├── go.work
├── node_modules/
└── package.json
```

The relevant content of my `.air.toml` file looked like this when i came across this issue and tried to enable watching on the libs for the backend:
```toml
root = "."
tmp_dir = "apps/backend/tmp"
testdata_dir = "apps/backend/testdata"

[build]
bin = "apps/backend/tmp/main.exe"
cmd = "go build -o apps/backend/tmp/main.exe apps/backend/cmd/main.go"
exclude_dir = ["assets", "tmp", "vendor", "testdata"]
exclude_file = []
exclude_regex = ["_test.go"]
exclude_unchanged = false
follow_symlink = false
full_bin = ""
include_dir = ["libs\\backend"]
include_ext = ["go"]
```
(i know that i need to add `apps\\backend` there as well, but thats the state of the file when i stumbled across this issue)

Notice where everywhere in the config file, forward-slash notation can be used even when running Windows, but not for `include_dir`.
Would i define it like this `include_dir = ["libs/backend"]`, air would ignore the directory, noticeable by its log output:
```
λ air -c apps/backend/.air.toml

  __    _   ___
 / /\  | | | |_)
/_/--\ |_| |_| \_ v1.61.7, built with Go go1.24.1

!exclude api
!exclude apps
!exclude database
!exclude dist
!exclude docker
!exclude libs\backend
!exclude libs\frontend
!exclude node_modules
building...
running...
```
You can see, that pretty much my whole project (i stripped some dirs from the graphic above to make it not so verbose) is excluded. On that note, not sure if its just me but prefixing `exclude` with an exclamation mark (-> `!exclude`) initially caused a lot of confusion on my end. Im not sure why that notation was chosen, but it's an odd choice and causes wrong assumptions about the program behavior.

[...]

PR #866, opened 2026-01-19
feat: add build-specific platform overrides

xiantang:
close https://github.com/air-verse/air/issues/279


xiantang:
also need to consider this case https://github.com/air-verse/air/issues/767

Issue #849, opened 2026-01-10
Preload .env files


sirkostya009:
#### Is your feature request related to a problem? Please describe.
Air doesn't have a feature for setting and expanding (preloading) files from a .env file. This is frustrating as all of my apps already use air and I'd like to not not have code for parsing .env files in my binary as it only runs in development.

#### Describe the solution you'd like
Setting and expanding environment variables using Go standard library functions. Should default to a local .env file, but can be of course configured.

#### Describe alternatives you've considered
Manual preload in main via godotenv

---

Sorry if this was discussed before, I couldn't find any issues or related discussions. There's two PRs open for this feature and I hope they get merged soon!

xiantang:
Hello, it's really good feature, but still have sth need to confrim. so for air there are two stage, build and run. Which stage will you apply your .env

xiantang:
btw PRs are welcome

mehdignu:
currently i use, in `.air.toml`
```
bin = "./tmp/someApp"
cmd = "go build -o ./tmp/someApp ./cmd/api"
...
full_bin = "ENV_PATH=../project/.env ./tmp/someApp"
```

and it works for like this in my monorepo `main.go`
```

	envPath := os.Getenv("ENV_PATH")

	if envPath != "" {
		if err := godotenv.Load(envPath); err != nil {
			log.Println("env file not found at:", envPath)
		} else {
			log.Println("env file loaded successfully")
		}
```

xiantang:
> currently i use, in `.air.toml`
> 
> ```
> bin = "./tmp/someApp"
> cmd = "go build -o ./tmp/someApp ./cmd/api"
> ...
> full_bin = "ENV_PATH=../project/.env ./tmp/someApp"
> ```
> 
> and it works for like this in my monorepo `main.go`
> 
> ```
> 
> 	envPath := os.Getenv("ENV_PATH")
> 
> 	if envPath != "" {
> 		if err := godotenv.Load(envPath); err != nil {
> 			log.Println("env file not found at:", envPath)
> 		} else {
> 			log.Println("env file loaded successfully")
> 		}
> ```

yes we know there are two ways to pass ENVs.

1. `export XXXX=XXX air`
2. define in full_bin

im considering to make air respect .env

silverwind:
This should be a opt-in feature via configuration which files are sourced in which order. Many tools support multiple files, for example vite: https://vite.dev/guide/env-and-mode#env-files

PR #867, opened 2026-01-19
Implement .env file preload

xiantang:
closes https://github.com/air-verse/air/issues/849

so my implementation uses godotenv to load and reload the .env file making the loaded env available for both build and run steps.

also thinking maybe there should be a separate config option for build and run respectively?

does not override existing global env values


sirkostya009:
@xiantang but what about these prs? https://github.com/air-verse/air/pull/751 https://github.com/air-verse/air/pull/598

there's also https://github.com/air-verse/air/pull/47 but the implementation is somewhat different

Issue #869, opened 2026-01-19
No console logs when running on windows


BryceBarbara:
**Describe the bug**  
When running an application with air on Windows, no application logs are seen.

**To Reproduce**  
- [X] Provide a minimal reproducible example: https://github.com/air-verse/air-reproducible-example/pull/6

**Expected behavior**  
I'd expect to see the application logs (stdout, stderr) in the terminal.

**Screenshots**  
<img width="604" height="201" alt="Image" src="https://github.com/user-attachments/assets/a69961b0-96aa-40c9-a3f2-6c5547dbe3f4" />

**Desktop:**  
- OS: Windows 11 25H2
- Air Version: v1.64.1 (Go v1.25.5)

xiantang:
ok tks for reporting

Ikuradachi:
any fix? 👀

xiantang:
please use v1.64.0 first

xiantang:
should be fixed on latest version

PR #870, opened 2026-01-20
Fix Windows logging attachment

xiantang:
close https://github.com/air-verse/air/issues/869


xiantang:
close https://github.com/air-verse/air/issues/869

PR #879, opened 2026-02-01
Added Homebrew installation instructions to README.md

alesker:
Air is now available in the Homebrew core, so listing it as the option in installation instructions


xiantang:
thank you so much let me check it out

Issue #882, opened 2026-02-23
Update dockerfile to go 1.26


gouthamve-but-enterprise:
We just updated our go.mod to require `1.26` and our dev setup broke with:

```
go build ......

go: go.mod requires go >= 1.26 (running go 1.25.4; GOTOOLCHAIN=local)
```

We use the `air` docker image and it's currently at `1.25`: https://github.com/air-verse/air/blob/master/Dockerfile#L15

Can we update it to `1.26`? I am happy to send a PR :)

mar4uk:
yeah, I think we need to update to 1.26. Please send PR if you can 🙂

yetanotheraryan:
@xiantang @mar4uk  Can I do this, by updating the base image in the Dockerfile without modifying go.mod ? which you have mentioned in the PR for @gouthamve-but-enterprise ?

xiantang:
> [@xiantang](https://github.com/xiantang) [@mar4uk](https://github.com/mar4uk) Can I do this, by updating the base image in the Dockerfile without modifying go.mod ? which you have mentioned in the PR for [@gouthamve-but-enterprise](https://github.com/gouthamve-but-enterprise) ?

yes pls, PR is welcome

PR #883, opened 2026-02-24
Update golang to 1.26

gouthamve-but-enterprise:
fixes https://github.com/air-verse/air/issues/882

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,7 +19,7 @@ jobs:
         id: go
         uses: actions/setup-go@v5
         with:
-          go-version: ^1.26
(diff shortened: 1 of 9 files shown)


xiantang:
Thanks for the PR! However, we'd prefer not to bump the Go version requirement in go.mod at this time.

Raising the minimum Go version would break go install for users on older Go versions. As a development tool, air should maintain broad compatibility across Go versions.

To address #882, we can update the base image in the Dockerfile without modifying go.mod. Could you adjust the PR to only change the Dockerfile and keep go.mod as-is?

gouthamve-but-enterprise:
@xiantang Done, thanks for the review :)

PR #908, opened 2026-05-20
fix: rewatch files after atomic saves

### LINKED ISSUE #496: Files in `root_dir` not watched properly when using `include_dir`
air version: 1.49.0

I would like to be able to watch all files with specific extensions in a set of directories as well as some files in `root_dir`. I have an example module layout like this:

```
$ tree
.
├── air.toml
├── go.mod
├── main.go
├── mypkg
|   └── exec.go
└── otherpkg
    └── file.go
```

For example, I would like to watch all `.go` files in `mypkg`, `main.go`, and `go.mod`. I first tried the following config:
```
include_dir = ["mypkg"]
include_ext = ["go"]
include_file = ["go.mod"]
```

I ran this in Docker with the following command:
```
docker run --rm -v $PWD:/src -w /src cosmtrek/air:v1.49.0 -d -c /src/air.toml
```

With this config, changes to `mypkg/exec.go` were detected fine and triggered a rebuild, no changes to `main.go` were detected, and some changes to `go.mod` were detected, but it did not trigger a rebuild. For `go.mod` it seems to not process all the same file events as for `mypkg/exec.go`.

Specifically, the *first* save to `go.mod` produces these file events:
```
[17:09:38] event: REMOVE        "/src/go.mod"                                                                                                                                                              
[17:09:38] event: CHMOD         "/src/go.mod"
```
and successive changes to `go/mod` do not produce *any* file events.

For `mypkg/exec.go`, many more file events are detected when the file is saved, and they are observed on every save:
```
[17:16:16] event: RENAME        "/src/mypkg/exec.go"
[17:16:16] event: WRITE         "/src/mypkg/exec.go"
[17:16:16] event: CREATE        "/src/mypkg/exec.go"
[17:16:16] mypkg/exec.go has changed
[17:16:16] mypkg/exec.go has changed
[17:16:16] event: CHMOD         "/src/mypkg/exec.go"
[17:16:17] flushing events
[17:16:17] mypkg/exec.go has changed
```

So it seems like I could get `main.go` to be watched by adding it to `include_file`. But the watching provided by this configuration seems incomplete as file events are not always detected and do not trigger rebuilds, as seen for the example with `go.mod` above. I could also remove the `include_dir` setting and then things mostly work as desired, but then it also watches the `otherpkg` directory, which is not desired. Maybe I could do that and then used `exclude_dir = ["otherpkg"]`? That seems to work, but it doesn't scale well when there are lots of directories and you really only want to include one or a few of them. I also adding `"."` to `include_dir`, which seems to have no effect.

Maybe this is a misunderstanding on my part of how the various parameters for including and excluding things are supposed to be used. For example, maybe I need to include `"mod"` in the `include_ext` to have `go.mod` watched properly? It's not clear (I did try it and it didn't help). These could *really* use some documentation because it's not clear at all how they interact, which take precedence, which can work together, etc.

To summarize:
* There is some strange behavior with file watching when using `include_file` in combination with `include_dir` where incomplete sets of file events are generated and the binary rebuild is not triggered.
* The interactions between the various `include_*` and `exclude_*` parameters are confusing and could be better documented.

xiantang:
## Summary
Fix watcher handling for editors that save files atomically by rename/remove. Air now treats rename events as valid and re-adds watched files after rename/remove events.
close https://github.com/air-verse/air/issues/496

## Changes
- Treat fsnotify rename events as valid file changes.
- Re-add watched files after rename or remove events, with error logging.
- Update event helper tests for rename and remove handling.

## Testing
- go test ./...
- make check

## Screenshots
N/A

## Checklist
- [x] Tests pass
- [ ] Documentation updated
- [x] No breaking changes

### DIFF
--- a/runner/engine.go
+++ b/runner/engine.go
@@ -322,8 +322,10 @@ func (e *Engine) watchPath(path string) error {
 				}
 				// Rewatch the file if the editor is using atomic saving.
 				if renameOrRemoveEvent(ev) {
-					if err := e.watcher.Add(ev.Name); err != nil {
-						e.watcherLog("error rewatching file: %v", err)
+					if e.checkIncludeFile(path) {
+						if err := e.watcher.Add(ev.Name); err != nil {
+							e.watcherLog("error rewatching file: %v", err)
+						}
 					}
 				}
 				e.watcherDebug("%s has changed", e.config.rel(ev.Name))
--- a/runner/engine.go
+++ b/runner/engine.go
@@ -322,7 +322,7 @@ func (e *Engine) watchPath(path string) error {
 				}
 				// Rewatch the file if the editor is using atomic saving.
 				if renameOrRemoveEvent(ev) {
-					if e.checkIncludeFile(path) {
+					if e.checkIncludeFile(ev.Name) {
 						if err := e.watcher.Add(ev.Name); err != nil {
 							e.watcherLog("error rewatching file: %v", err)
 						}
--- a/runner/engine.go
+++ b/runner/engine.go
@@ -320,6 +320,12 @@ func (e *Engine) watchPath(path string) error {
 				if !e.isIncludeExt(ev.Name) && !e.checkIncludeFile(ev.Name) {
 					break
 				}
+				// Rewatch the file if the editor is using atomic saving.
+				if renameOrRemoveEvent(ev) {
+					if err := e.watcher.Add(ev.Name); err != nil {
+						e.watcherLog("error rewatching file: %v", err)
+					}
+				}
 				e.watcherDebug("%s has changed", e.config.rel(ev.Name))
 				e.eventCh <- ev.Name
 			case err := <-e.watcher.Errors():
--- a/runner/util.go
+++ b/runner/util.go
@@ -303,11 +303,17 @@ func isDir(path string) bool {
 func validEvent(ev fsnotify.Event) bool {
 	return ev.Op&fsnotify.Create == fsnotify.Create ||
 		ev.Op&fsnotify.Write == fsnotify.Write ||
-		ev.Op&fsnotify.Remove == fsnotify.Remove
+		ev.Op&fsnotify.Remove == fsnotify.Remove ||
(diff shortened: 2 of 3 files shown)


xiantang:
@codex review

xiantang:
@codex review

xiantang:
@codex review

xiantang:
@silverwind if you interested

silverwind:
Verified the atomic-save premise — it's correct. Air's default watcher (`filenotify.NewEventWatcher` → fsnotify) binds to the inode (inotify on Linux, kqueue/vnode on macOS), so an atomic save relinks the path to a new inode and the old watch goes deaf — exactly the `REMOVE`-then-silence seen in the issue. Re-adding follows the new inode, so the approach is sound. `include_dir` is unaffected because a directory's inode is stable across a child's atomic save.

Confirmed empirically on macOS that the fix re-fires on every real save pattern (rename-into-place, remove+create, rename-away+write); without the re-add the 2nd save onward is silent. The earlier Codex P1/P2 points are addressed by the latest commit (`ev.Name` instead of the captured `path`, and the `checkIncludeFile` gate scoping the re-add to `include_file`).

Two optional, non-blocking follow-ups:

1. `watcher.Add` failure is only logged, never retried. There's a narrow window where the path is momentarily absent (rename-away before recreate) and the re-add fails, leaving the file permanently unwatched. Realistic async timing didn't hit it, but a slow/Docker bind-mount FS could — a retry, or watching the parent dir per fsnotify's own recommendation, would close it.
2. No end-to-end regression test — the tests cover the boolean helpers but not the rewatch behavior itself.

_Review written with Claude Opus 4.7._

silverwind:
Looks good actually, premise is correct as per above review from Claude.

xiantang:
> Looks good actually, premise is correct as per above review from Claude.

thanks let me merge it

silverwind:
Still recommended to address the "non-blocking follow-ups" mentioned above.

xiantang:
> Still recommended to address the "non-blocking follow-ups" mentioned above.

Sure

PR #826, opened 2025-11-17
deprecated: --build.bin --build.args_bin

xiantang:
close https://github.com/air-verse/air/issues/824

PR #912, opened 2026-05-21
fix: honor --build.bin when no config file exists

hiendv:
## Summary

Fixes `--build.bin` being silently ignored when Air runs without a config file.

When no `.air.toml` or `.config/air.toml` exists, Air pre-populated `build.entrypoint` from the default `build.bin` before CLI flags were applied. `--build.bin tmp/server` updated `Build.Bin`, but runtime still preferred the stale default entrypoint and tried to run `./tmp/main`.

## Changes

- Stop pre-populating `Build.Entrypoint` when falling back to built-in defaults.
- Track whether config came from a file so plain `air` does not emit a false `build.bin` deprecation warning.
- Still warn when `build.bin` is explicitly used from TOML or `--build.bin`.
- Add regression tests for no-config `--build.bin`, including the `--build.cmd ... --build.bin ...` repro.

## Semver Note

`--build.bin` is deprecated but still accepted, so users can reasonably expect it to keep working throughout `v1.x`.

If Air removes `--build.bin`, the semver-safe path is to keep it functional with warnings in `v1.x`, then remove or reject it in `v2.0.0`. This PR keeps the deprecated flag usable while it remains accepted.

## Context

This is related to the migration from `build.bin` to `build.entrypoint` in earlier changes such as #826 and #838. This PR does not change that migration path; it only fixes the no-config case where the deprecated flag was accepted but shadowed by the default entrypoint.

## Validation

```sh
go test ./runner -run '^TestDefaultPathConfig|^TestDefaultPathConfigWithInvalidTOML|^TestDefaultPathConfigDoesNotPrepopulateEntrypoint|^TestConfigRuntimeArgs|^TestFlag|^TestConfPreprocess|
^TestEntrypoint|^TestInitConfigWithoutConfigDoesNotWarnDeprecatedBin|^TestWarnDeprecatedBin$' -count=1 -v

### DIFF
--- a/runner/config.go
+++ b/runner/config.go
@@ -202,7 +202,7 @@ func (t sliceTransformer) Transformer(typ reflect.Type) func(dst, src reflect.Va
 
 // InitConfig initializes the configuration.
 func InitConfig(path string, cmdArgs map[string]TomlInfo) (*Config, error) {
-	ret, err := initConfigWithoutPreprocess(path)
+	ret, fromFile, err := initConfigWithoutPreprocess(path)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


xiantang (approved):

xiantang:
LGTM

Issue #915, opened 2026-06-25
air -v output defies conventions


jsshapiro:
**Describe the bug**  
Output of `air -v` and `air version` sent to wrong channel and not suitable for parsing

**To Reproduce**  
- Run "air -v > /dev/null". Observe that you still get version output.
- Then notice it's not in any reasonably parsable form.

**Expected behavior**  
The strong UNIX convention is that output of "cmd -v" or "cmd version" goes to stdout, on the logic that it is the intended output of the command given, and because one of the important use cases is parsing by other tools to check what version is currently installed. The air command incorrectly sends this output to stderr

Compare what "go version" does to see how this should work.

In addition to sending the version information to the wrong place, the output is conflated with the "A I R" logo, which is cute but not helpful.

When invoked as "air version", air incorrectly proceeds to wait, when the version command should emit the version and exit. The only thing that is requested by "air version" is to print the version number, which is what "air -v" (correctly) does.

**Desktop (please complete the following information):**  
- OS: MacOS (all versions), Linux (all versions). Observed on MacOS 26.5.1
- Air Version: 1.65.3

***Recommended Fix**

1. Send output to stdout rather than stderr
2. Emit the actual version information on the line _below_ the logo so that it can be extracted using grep and sed.

xiantang:
close, cuz already fixed

PR #916, opened 2026-06-27
fix: send version output to stdout on `air -v`

sysrecall:
Related to #915 

## Changes
- Send version output to stdout instead of stderr on `air -v`
- Keep version output suitable for parsing by scripts


xiantang (approved):
looks good, thanks

PR #937, opened 2026-08-25
fix: stream and bounds build command output

### LINKED ISSUE #932: 【BUG】Build command output is fully buffered before being printed
## Summary

`runCommandCopyOutput()` reads the entire build command stdout into memory before writing it to the terminal. For verbose build commands, this can cause unnecessarily high memory usage and delayed output.

## Affected Code

In `runner/engine.go`, `runCommandCopyOutput()` reads all stdout at once:

```go
stdoutBytes, _ := io.ReadAll(stdout)
_, _ = io.Copy(os.Stdout, strings.NewReader(string(stdoutBytes)))
```

This path is used by `building()` when running `e.config.Build.Cmd`.

## Safe Validation Evidence

Tested locally against commit `b487a2c59460eea59cee389dfb36152eab5b6eb0`.

A local package test used a build command that produced large stdout. The test confirmed that the output is accumulated in full before being copied to stdout:

```text
=== RUN   TestReproH7H8_UnboundedCommandOutput
    repro_high_test.go:121: H7/H8 = TRUE POSITIVE: seq 1 100000 = 588895 bytes. engine.go:687 io.ReadAll(stdout) reads ALL with NO LimitReader
--- PASS: TestReproH7H8_UnboundedCommandOutput (0.00s)
PASS
ok  	github.com/air-verse/air/runner	(cached)
```

The current `master` branch still contains the same code path at `runner/engine.go`.

## Expected Behavior

Build command output should be streamed or bounded so Air does not need to keep the entire stdout content in memory.

## Actual Behavior

Air reads the complete stdout stream into a byte slice before printing it.

## Impact

Very verbose build commands can cause high memory usage and delayed terminal output during rebuilds.

## Suggested Fix

Stream stdout directly while preserving enough output for the build failure message, or use a bounded buffer for the retained output.

Ali-Hasan-Khan:
Fixes: https://github.com/air-verse/air/issues/932

## Summary
The proxy build-failure overlay always showed an empty Output section:
startCmd created StdoutPipe/StderrPipe but immediately overwrote
Cmd.Stdout/Stderr with terminal fds, orphaning the pipes so
io.ReadAll captured "" every time.

## Changes
- startCmdWithOptions(capture bool) across linux/unix/windows; legacy
  startCmd untouched for pre/post cmds 
- runCommandCopyOutput streams stdout+stderr live, capturing a merged
  bounded 64KiB tail (new tailBuffer) for BuildFailedMsg
- tests: TestTailBuffer + bounded-capture integration test
- includes a separate first commit fixing pre-existing -race flakes that
  blocked verification of this change

## Testing
- go test ./... -race -timeout=3m ✅ 
- make check scope=all ✅
- Manual e2e: forced compile failure now renders compiler output in the overlay; seq 1 200000 stays ≤64KiB with tail retention while streaming live.

## Working Proof (for proxy overlay missing output)
- empty overlay output section(existing):
Screenshot From 2026-08-25 13-28-40
- showing terminal output/compile error in output section(fixed/expected):
image

### DIFF
--- a/runner/engine.go
+++ b/runner/engine.go
@@ -674,25 +674,32 @@ func (e *Engine) runCommand(command string) error {
 }
 
 func (e *Engine) runCommandCopyOutput(command string) (string, error) {
-	// both stdout and stderr are piped to the same buffer, so ignore the second
-	// one
-	cmd, stdout, _, err := e.startCmd(command)
+	cmd, stdout, stderr, err := e.startCmdWithCapture(command)
 	if err != nil {
 		return "", err
 	}
 	defer func() {
 		stdout.Close()
+		stderr.Close()
 	}()
 
-	stdoutBytes, _ := io.ReadAll(stdout)
-	_, _ = io.Copy(os.Stdout, strings.NewReader(string(stdoutBytes)))
+	var captured tailBuffer
+
+	stderrDone := make(chan struct{})
+	go func() {
+		defer close(stderrDone)
+		_, _ = io.Copy(io.MultiWriter(os.Stderr, &captured), stderr)
+	}()
+
+	_, _ = io.Copy(io.MultiWriter(os.Stdout, &captured), stdout)
+	<-stderrDone
 
 	// wait for command to finish
 	err = cmd.Wait()
 	if err != nil {
-		return string(stdoutBytes), err
+		return captured.String(), err
 	}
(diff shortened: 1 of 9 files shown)


Ali-Hasan-Khan:
Hi, @xiantang @GG-Feng please review

PR #941, opened 2026-09-08
chore: vendor filenotify to drop large hugo dependency

NathanBaulch:
I recently upgraded all modules in my project and stumbled on this error when trying to use air:

```
➜ go build github.com/air-verse/air
# github.com/gohugoio/hugo/common/hmaps
../../../../go/pkg/mod/github.com/gohugoio/hugo@v0.165.0/common/hmaps/maps.go:148:15: undefined: glob.Glob
```

I was surprised to see air has a dependency on a website framework!
The above error will be resolved in time once hugo releases [their fix](https://github.com/gohugoio/hugo/pull/15276) and air upgrades, however I think this is a good opportunity to remove this huge dependency and vendor the tiny `gohugoio/hugo/watcher/filenotify` sub-package instead (which itself was adapted from `moby/moby/pkg/filenotify`).

* Removing hugo shrinks `go.sum` by over 80%.
* The filenotify sub-package is very stable - no bug fixes in the last 3 years.
* Apache-2.0 license is compatible with air's GPLv3 license.


NathanBaulch:
Also, a 30% reduction in binary size.

| Build | before | after | Δ |
|---|---|---|---|
| Default (`go build`) | 18.16 MB | 12.63 MB | −5.53 MB (−30.5%) |
| Stripped (`-ldflags="-s -w"`) | 12.07 MB | 8.37 MB | −3.70 MB (−30.6%) |

